PR #10, opened 2017-12-05
Fix namedtemporaryfile

sdementen:
fix issue identical to https://stackoverflow.com/questions/18903069/how-can-i-read-namedtemporaryfile-in-python on windows (probably does not happen linux)


michaelboulton:
Didn't realise people would be using this on windows! Thanks for this

PR #13, opened 2017-12-09
WIP: python 2.7 support

benhowes:
[Update]: Added support for running in python 2.7.

I've attempted to patch python 3 methods using the future library as well as a few custom ones in `util/python_2_util` so that there are very few changes in the bulk of the code, other than `raise ... from ...` becoming `rasie_from(...)`.

### DIFF
--- a/tavern/response.py
+++ b/tavern/response.py
@@ -4,6 +4,9 @@ import textwrap
 import logging
 import copy
 
+from future.standard_library import install_aliases
+install_aliases()
+
 from urllib.parse import urlparse, parse_qs
 
 from .schemas.extensions import get_wrapped_response_function
@@ -61,7 +64,7 @@ class TResponse:
         else:
             return "<Not run yet>"
 
-    def _adderr(self, msg, *args, e=None):
+    def _adderr(self, msg, e=None, *args):
         if e:
             logger.exception(msg, *args)
         else:
--- a/tavern/response.py
+++ b/tavern/response.py
@@ -64,7 +64,9 @@ class TResponse:
         else:
             return "<Not run yet>"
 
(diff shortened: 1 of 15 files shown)


tavern/response.py
```diff
@@ -61,7 +64,9 @@ def __str__(self):
         else:
             return "<Not run yet>"
 
-    def _adderr(self, msg, *args, e=None):
+    def _adderr(self, msg, *args, **kwargs):
+        e = kwargs.get('e')
```
benhowes: I had to do this because the argument parsing works a little differently in 2.7. Of 4/5 ways I tried to get it going, this is the only one which works over 2&3 that I've found

tox.ini
```diff
@@ -1,10 +1,13 @@
 [tox]
-envlist = py33,py34,py35,py36,pyflakes
+envlist = py27,py33,py34,py35,py36,pyflakes
 skip_missing_interpreters = true
 
 [testenv]
 commands =
     {envpython} -bb setup.py test {posargs}
+deps =
```
benhowes: I had trouble running it locally without the deps here, but not sure if I was doing something wrong.

tavern/util/python_2_util.py
```diff
@@ -0,0 +1,41 @@
...
+        ]
+
+    _get_dict = c.pythonapi._PyObject_GetDictPtr
+    _get_dict.restype = c.POINTER(c.py_object)
+    _get_dict.argtypes = [c.py_object]
+
+    def get_dict(object):
+        return _get_dict(object).contents.value
+
+    get_dict(str)['format_map'] = format_map
```
michaelboulton: Instead of doing this it might be easier to just change the one use of `format_map` to use `format`

PR #28, opened 2018-02-08
Feature/constant session

### LINKED ISSUE #23: Implementing the Cookies Variable?
I'm having some trouble implementing a session using tavern and am a little confused as to how that would work.  Normally in python I would create a session object as follows:

```
with reqeusts.Session() as sesh:
  data = { "username" : "myName", "password" : "myPW"}   
  sesh.get(url,data = data, ... )
  etc.
```

and then after this just making all of the necessary calls as needed using sesh, however I'm just not sure how to approach this within tavern.  My first thought was to create an external function that would act like the Sessions function and grab the cookies, auth, and something else (can't remember off the top of my head right now) and then simply copy and paste that code to each different test.  My problem though was that the keywords for auth and for cookies has not yet been implemented.

My main questions are

1) Is there a way to currently implement a constant session with the way that tavern is right now?
2) If not, how could we implement this and what would need to be done in order to get this type of functionality?

Here is the code that I wanted to implement but cannot mainly because cookies has not yet been implemented:

```
# test_sz.tavern.yaml

---
test_name: Testing *** Module

stages:
- name:  %DOMAIN%/api/authenticate POST
  request:
    url: %DOMAIN%/api/authenticate
    method: POST
    data: 
      password: ***
      username: ***
      thirdPartyLogin: false
      forcelogin: false
    headers:
      content-type: application/x-www-form-urlencoded
      accept: application/json 
    verify: false
  response:
    save:
      $ext:
        function: helperFuncsTavern:getCookie
    status_code: 200

- name: Testing %DOMAIN%/api/search/category GET
  request:
    url:  %DOMAIN%/api/search/category 
    method: GET
    headers:
      accept: application/json
    verify: false
    cookies: cookies.seshCookie
  response:
    status_code: 200
```

Here is the code from the getCookie external function:

```
def getCookie(response):
""" Grabs the seshCookie cookie from the response so that
    we can reuse it throughout the rest of the session.  """
    return {"cookies" : {"seshCookie" : response.cookies["seshCookie"]}}
```

Appreciate the help and hope to hear from you soon!

Edit: Looking at the code again I found that there is actually support for the auth keyword, so ignore that part from above about auth.

michaelboulton:
Use a constant session for each test so that cookies are retained between stages.

Also add an example of a server that uses cookies + tavern test file that shows how cookies are reused implicitly.

closes #23


benhowes:
@michaelboulton This looks great to me! I think we really need some updated documentation to explain how the cookies functionality works though.

WhileLoop:
Just tried this. Seems the cookie is being included in subsequent requests even if its not specified in  response/cookies. Personally I prefer this behavior over having to explicitly list each cookie but from what I understand it doesn't match the documentation update.

WhileLoop:
This is working for me. Would love to see this merged and shipped.

PR #29, opened 2018-02-16
add regex validation function

WhileLoop:
add a regex validation helper function and a regex example

### DIFF
--- a/tavern/testutils/helpers.py
+++ b/tavern/testutils/helpers.py
@@ -83,17 +83,24 @@ def validate_pykwalify(response, schema):
     with wrapfile(response.json()) as rdump, wrapfile(schema) as sdump:
         verify_generic(rdump, sdump)
 
-def validate_regex(response, expression, expected_matches):
+def validate_regex(response, expression):
     """Make sure the response body matches a regex expression
 
     Args:
         response (Response): reqeusts.Response object
         expression (str): Regex expression to use
-        expected_matches (list): list of strings to compare to matched groups
+    Returns:
+        dict: dictionary of regex: boxed full_match, and list of matched groups
     """
-    matched = re.match(expression, response.text).groups()
-    for i in range(len(expected_matches)):
-        if not expected_matches[i] == matched[i]:
-            raise Exception(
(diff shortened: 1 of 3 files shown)


WhileLoop:
Is it possible to apply functions to anything other than response body?

benhowes:
@WhileLoop Thanks, this is a great addition!

I suspect that you might be able to get away without the group matching aspect of this for the most part? Unless I'm missing a case where you need to be able to match the groups individually?

WhileLoop:
I made the group matching useful by allowing them to be saved and reused.

benhowes:
@WhileLoop Sorry for the delay in getting back to you on this. I've been thinking about it a bit and wondering if a nicer way of doing this could be to use "named groups" in regexes. This could mean that we automatically save the contents of the named groups to the name if provided. Un-named groups could be dropped. The re module provides [this](https://docs.python.org/3.6/library/re.html#re.match.groupdict) which would make it pretty easy to implement too.

What do you think?

tavern/testutils/helpers.py
```diff
@@ -90,17 +90,12 @@ def validate_regex(response, expression):
...
     assert match
 
-    full_match = match.group(0)
-    groups = [group for group in match.groups()]
     return {
-        "regex": Box({
-            'full_match': full_match,
-            'groups': groups
-        })
+        "regex": Box(match.groupdict())
```
benhowes: I reckon they might be okay without being keyed on the dict, i.e. they could just be top level variables? so you can just return the `Box(match.groupdict())`

benhowes:
I think this is looking good @WhileLoop! I think @michaelboulton should have a look too and possibly a few tests of the functionality can be added.

tavern/testutils/helpers.py
WhileLoop: This follows the pattern of the other helpers. Also returning the matches as top level variables increases chance some other existing variable being overwritten inadvertently.

michaelboulton:
The only (minor) issue with this is that it forces people to use named match groups, but I actually like the named groups

Issue #26, opened 2018-02-07
Publishing MQTT message(s) before commencing test


benhowes:
How do you think this should work @michaelboulton? Can you come up with something and implement it with some examples in the docs site?

michaelboulton:
Might be best if we allow either requests or responses to be publishing/expecting an mqtt message. 

```yaml
---

test_name: Make sure posting sends mqtt message

stages:
  - name: step 1 - post message trigger
    request:
      url: "{host}/send_mqtt_message"
      json:
        device_id: 123
        payload: "hello"
      method: POST
      headers:
        content-type: application/json
    mqtt_response:
      topic: /device/123
      payload: "hello"

---

test_name: Make sure sending mqtt message triggers another message

stages:
  - name: step 1 - post message trigger
    mqtt_request:
      topic: /device/123
      payload: "ping"
    mqtt_response:
      topic: /device/123/response
      payload: "pong"
      message_timeout: 5

```
- Every test would result in a clean mqtt session (similar to #23)
- All messages between the request and the response would be queued
- Not receiving an expected message within a given timeout would be an error
- Receiving any other messages would either be an error or a warning - configurable?
- Either automatically subscribe to any topics in `mqtt_response` or have a list of topics to subscribe to for each test/stage

michaelboulton:
As discussed:

- mqtt_request renamed to mqtt_publish
- Can only have `mqtt_publish` OR `request`, not both. If there isn't a `request`, can't have a `response`.
- If there is a `request`, there needs to be a `response`
- Run in this order:
  1. mqtt_publish
  2. request
  3. response
  4. mqtt_response

example:

```yaml
---

test_name: Make sure posting sends mqtt message

stages:
  - name: step 1 - post message trigger
    request:
      url: "{host}/send_mqtt_message"
      json:
        device_id: 123
        payload: "hello"
      method: POST
      headers:
        content-type: application/json
    response:
      status_code: 200
      body:
        message: "success"
      headers:
        content-type: application/json
    mqtt_response:
      topic: /device/123
      payload: "hello"
      timeout: 5
```

ie, we send a http request, which returns a 200, and also triggers an mqtt message published on /device/123 within 5 seconds

michaelboulton:
example https://github.com/taverntesting/tavern/blob/feature/mqtt/example/mqtt/test_mqtt.tavern.yaml

michaelboulton:
[...]

Issue #20, opened 2018-01-24
Value mismatch error when values seem to be the 'same'


casahome2000:
Command
```
$ tavern-ci --debug some_file.tavern.yaml
```

Response in Terminal
```
Value mismatch: '159295' vs '159295'
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/tavern/response.py", line 210, in _validate_block
    assert actual_val == expected_val
AssertionError
```

Test to save `user_id` from response
```
  - name: login
    request:
      url: some_test_url.com/login
      json:
        username: someUser
        password: somePassword
        grant_type: password
      method: POST
      headers:
        content-type: application/json
        accept: application/json
    response:
      status_code: 200
      body:
        token_type: bearer
        scope: user
        expires_in: 604800
      headers:
        content-type: application/json; charset=utf-8
      save:
        body:
          access_token: access_token
          refresh_token: refresh_token
          user_id: user_id
```

Test to use `user_id` in subsequent test
```
  - name: create_book
    request:
      url: some_test_url.com/books
      json:
        title: Some title of a book
      method: POST
      headers:
        content-type: application/json
        accept: application/json
        Authorization: "Bearer {access_token:s}" #working to reuse variable saved
    response:
      status_code: 201
      body:
        user_id: "{user_id}"
      headers:
        content-type: application/json; charset=utf-8
```

michaelboulton:
I'm guessing this is because it's an integer and user_id is being implicitly converted to an int at some point.

Because it might be desirable to actually have the distinction between an integer and a string being returned I'm not sure what the best thing to do is to fix this - it should at least print something more useful like

```
Value mismatch: '159295' (type: int) vs '159295' (type: str)
```

benhowes:
@michaelboulton This has been released right?

Birne94:
Currently, values from variables can only be passed by using string formatting (`user_id: "{user_id}"`). This loses any information about the original data type.

We should come up with another way of passing variables as they are. This might also be useful for passing whole json-blocks as data. String-formatting should stay as it has its own use case (e.g. `Authorization` header formatting).

Maybe a special token might be appropriate here. I am not sure which tokens are reserved in the yaml spec, so the symbol might need to be changed.

```yaml
user_id: $user_id
```

michaelboulton:
[...]

Issue #2, opened 2017-11-08
Make a pretty printer for errors


michaelboulton:
Something like the default pytest one:

https://github.com/pytest-dev/pytest/blob/master/_pytest/terminal.py#L126

We want it to print out the yaml that caused the failure, the line, the file, etc. Colourised would be nice of course.

michaelboulton:
Make it so each expected/request/verifier type has a log_fail which helpfully explains why it failed as well, or have some kind of pytest/user hook so it can do 'something' when it fails. At the moment it's essentially pointless

ahopkins:
This would be awesome. The names are certainly a help, but I have taken to providing a sort of serial number in the name field so that I can easily identify where in my yaml the problem is. It is all manual right now, so it would be helpful if there was an easier way to track down the source of a failure.

Awesome work on this by the way, and thanks for open sourcing it!

michaelboulton:
I started this with some help from the Pytest people: https://github.com/pytest-dev/pytest/issues/3586

The real issue is trying to present the code in a nice format and showing which errors were caused at which point. This involves getting the yaml for the stage, probably post-format, which is currently a bit difficult.

It might be easier to generate a big output log file and then just parse it.

michaelboulton:
Progress:

![err2](https://user-images.githubusercontent.com/3809868/41662813-d6d21d8c-7499-11e8-9eff-b22bd5c12566.png)

michaelboulton:
@benhowes I was going to merge this into master and enable the 'new' reporting with a flag (eg `--tavern-beta-rewrite-exceptions`). I think the real 'big' step for this issue is to add metadata to exceptions so that we can highlight the line in the stage that failed for each error (in the case where there is multiple errors). Some note would go in the documentation about this, but also how it's a beta feature that might get changed in future.

michaelboulton:
Also, print out the 'formatted' stage if possible alongside the raw input stage.

benhowes:
@michaelboulton would it make sense for highlighting multiple errors to be a future feature? The current functionality is a big improvement on where we are at the moment, so it may make sense to get that out in a release. Printing the formatted stage alongside the raw one would be a big help for debugging too

michaelboulton:
Highlighting multiple errors is definitely a fairly big thing so that can be left until a later release.

michaelboulton:
This is now in 0.13.0 - use the `--tavern-beta-new-traceback` flag to enable it

PR #30, opened 2018-02-16
Test MQTT messages

michaelboulton:
This is ready for reviewing but not merging

Added:

- `mqtt_publish` and `mqtt_response`. Specifications here https://github.com/taverntesting/tavern/issues/26#issuecomment-364094564 
- Add rudimentary plugin system (hardcoded to only use mqtt/rest requests at the moment) to clean up core.py
- Use ExitStack rather than a context manager for 'sessions' (including mqtt sessions) for ease of use + future plugins
- add MQTT server/listener example
- make checking expected input keys a utility function

Things that need thinking about:

- connection timeout in context manager for MQTT client - currently waits for a given number of seconds then raises an error if it can't connect. I think this seems OK?
- Currently doesn't raise an error if an unexpected message is received while waiting for another message
- automatically dumps yaml/json to a string if it's used as the payload - might want to properly specify this behaviour
- When publishing a message it subscribes to that topic and waits for it to arrive - this was to get around issues where the message would take longer to be broadcast than it would take to move onto the next test stage. We might want to keep this in but also add a 'sleep' stage type which just waits for a few seconds for something to be processed

Will add the docs once we've sorted all these issues out


example/mqtt/test_mqtt.tavern.yaml
```diff
@@ -0,0 +1,112 @@
...
+        content-type: application/json
+    response:
+      status_code: 200
+      body:
+        lights: "off"
+      headers:
+        content-type: application/json
+
+  - name: step 2 - send mqtt message saying that the lights are now on
+    mqtt_publish:
```
benhowes: I think we definately need a way to add a delay after publishing to a topic. perhaps `delay_after`?

example/mqtt/test_mqtt.tavern.yaml
```diff
@@ -0,0 +1,112 @@
...
+      headers:
+        content-type: application/json
+    response:
+      status_code: 200
+      body:
+        lights: "off"
+      headers:
+        content-type: application/json
+
+  - name: step 2 - send mqtt message saying that the lights are now on
```
benhowes: "sending" should be referred to as "publishing throughout"

example/mqtt/test_mqtt.tavern.yaml
```diff
@@ -0,0 +1,112 @@
...
+    client_id: tavern-tester
+
+stages:
+  - name: step 1 - ping/pong
+    mqtt_publish:
+      topic: /device/123/ping
+      payload: ping
+    mqtt_response:
+      topic: /device/123/pong
+      payload: pong
```
benhowes: What happens if more than one message is received on this topic, but the first N messages do not meet the validation?

example/mqtt/test_mqtt.tavern.yaml
```diff
@@ -0,0 +1,112 @@
...
+        content-type: application/json
+    response:
+      status_code: 200
+      body:
+        topic: "/device/123"
+      headers:
+        content-type: application/json
+    mqtt_response:
+      topic: /device/123
+      payload: "hello"
```
benhowes: How might we validate JSON or other encoded payloads?

tavern/mqtt.py
```diff
@@ -0,0 +1,228 @@
...
+        while time_spent < timeout:
+            t1 = time.time()
+
+            try:
+                msg = self._message_queue.get(block=True, timeout=timeout)
+            except Empty:
+                time_spent += timeout
+            else:
+                time_spent += time.time() - t1
+                if msg.payload != payload and msg.topic != topic:
```
benhowes: This if seems to fall into the `else` section if either the payload or the topic are correct, which does not appear to be the desired functionality. Looks that the and should be an `or`, or use the else block for the warning.

r.e. the comment, I don't think we should error on an unexpected message - I think we should just only pass when the exact right message is received.

tavern/mqtt.py
```diff
@@ -0,0 +1,228 @@
...
+        return False
+
+    def publish(self, topic, payload, *args, **kwargs):
+        """publish message using paho library
+        """
+        logger.debug("Publishing on '%s'", topic)
+        self._client.subscribe(topic, qos=1)
+
+        if isinstance(payload, (dict, list)):
+            payload = json.dumps(payload)
```
benhowes: Can't assume JSON, we should at least have a key in the yaml for this, even if only JSON is supported without an external plugin.

tavern/plugins.py
```diff
@@ -0,0 +1,154 @@
...
+    Args:
+        test_spec (dict): Spec for the test block
+
+    Returns:
+        dict: mapping of name: session. Session should be a context manager.
+    """
+
+    sessions = {}
+
+    if 1: # pylint: disable=using-constant-test
```
benhowes: Do you _have_ to load the HTTP stuff? presumably as of this branch, you could just be testing MQTT? If you do always need to load this, the `if 1` doesn't really make it easier to understand more than the comment.

tavern/printer.py
```diff
@@ -4,17 +4,19 @@
...
     logger.info(formatted)
 
 
 def log_fail(test, response, expected):
+    # pylint: disable=unused-argument
     # TODO print more info
     fmt = "FAILED: {:s} [{}]"
-    if response:
+    try:
         formatted = fmt.format(test["name"], response.status_code)
```
benhowes: Should doing this be handled by a method on the verifier or something? seems like it would be more flexible/extendable later?

tavern/mqtt.py
michaelboulton: we could just have a `json` key mutually exclusive with a `payload` key where json is automatically dumped

tavern/printer.py
michaelboulton: Definitely should be, the whole log_fail/log_pass should be reworked a lot

tavern/mqtt.py
michaelboulton: This should be fixed now

tavern/plugins.py
michaelboulton: Fixed

tavern/printer.py
michaelboulton: specifically this https://github.com/taverntesting/tavern/issues/2

example/mqtt/test_mqtt.tavern.yaml
michaelboulton: Now done using the same method as validating json responses from a server. We can rework that function (`BaseResponse.recurse_check_key_match`) to make this better in future (eg https://github.com/taverntesting/tavern/issues/20)

example/mqtt/test_mqtt.tavern.yaml
michaelboulton: They are ignored until the specified timeout has passed, at which point it fails the test

example/mqtt/test_mqtt.tavern.yaml
michaelboulton: fixed

example/mqtt/test_mqtt.tavern.yaml
michaelboulton: Is it best to have a delay_before or delay_after? Or both?

tavern/mqtt.py
michaelboulton: It now has a 'json' key alongside 'payload' which is automatically dumped/loaded as json and used as the payload.

PR #33, opened 2018-02-20
Added other verbs for HTTP requests.

### LINKED ISSUE #32: When will be support http method PATCH?
What version will be support this?

benhowes:
Closes #32.


michaelboulton:
Do all of these methods behave like a post request (eg, stuff sent in the post body)? There is a check to stop you sending a body in a GET request, wondering if there should be something like this for any of these methods (eg options) https://github.com/taverntesting/tavern/blob/f7448cd1f7bc5319d12a4f1ac90d95fcb63d8f24/tavern/request.py#L110-L112

The only reason all the verbs weren't included in the original response is I don't know about the semantics of them

benhowes:
@michaelboulton that's a good point. Most can, infact in some horrible cases (e.g. Elastic Search) it is possible to send a body on a GET request even. I will check up on it later.

benhowes:
As discussed, it's probably best if we don't enforce no bodies at all, but we can warn on `OPTIONS`, `GET,` and `HEAD` since they're not expected to have bodies (as per: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)

benhowes:
@michaelboulton

Issue #42, opened 2018-03-01
How to use a variable in the array?


vit-goncharov:
I'm trying to use my variable value  in the array but it doesn't work. E.g.:
```
json:
        settings: []
        testlist: ["{test_id:d}"]
```

vit-goncharov:
Send
```
json:
        settings: []
        testlist:
          - "{test_id:d}"
```

Print on server
['{test_id:d}']

benhowes:
Closed by #43

PR #43, opened 2018-03-01
Add variables support for arrays

Birne94:
This small PR adds the support for formatted strings/variables in arrays, see also #42.

```yaml
json:
  foo:
    - "{bar:s}"
```


benhowes:
@Birne94 Thanks, looks good to me! @michaelboulton can you merge if you're happy?

benhowes:
Since this is still open, let's also fix #45 in here.

Birne94:
Unless I misinterpreted #45, it should already work with the changes of this PR. I have added a test to verify this.

benhowes:
@Birne94 thank you so much for the contribution! My note earlier was actually for @michaelboulton, but thanks again!

Issue #8, opened 2017-11-27
Use more pytest features


michaelboulton:
Would be nice if we could use something like `@pytest.mark.parametrize` on tests. We can do stuff like this with pytest:

```python
def get_login_stage():
    return {
        "request": {
            ...
        }
        "response": {
            ...
        }
    }

@pytest.mark.parametrize("test_a_thing", (True, False))
def test_something(test_a_thing):
    stages = []

    stages.append(get_login_stage())

    if test_a_thing:
        stages.append(get_optional_stage())
    else:
        stages.append(get_other_stage())

    stages.append(get_final_stage()

    tavern.run_test(stages)
```
It would be nice if there was a way to do this in the yaml.

This can probably be done with some kind of custom YAML constructor (`!parametrize True False`...)

michaelboulton:
Stretch goal, but it would be good if we could put in generic pytest things in as well - eg, `xfail`, `skip`, relying on pytest fixtures, etc.

michaelboulton:
Everything except the fixtures/running code is done, which is described better in https://github.com/taverntesting/tavern/issues/98

PR #47, opened 2018-03-05
Add basic pytest fixture support [WIP]

Birne94:
# Work in progress

referencing #8 

This PR adds basic support for pytest fixtures. Fixtures specified in the test as well as autoloaded fixtures are evaluated and their results are available in the test.

basic example, adapted from the `simple` example from the repo:

```yaml
test_name: Make sure server doubles number properly
pytest:
  fixtures:
    # store result of fixture "number" in variable "input"
    input: number
    # store result of fixture "double" in variable "expected"
    expected: double

stages:
  - name: Make sure number is returned correctly
    request:
      url: http://localhost:5000/double
      json:
        number: "{input:d}"
      method: POST
      headers:
    response:
      status_code: 200
      body:
        # type information is lost here, so this fails.
        double: "{expected:d}"
```

**What works so far** 

* reference fixtures in the same way as variables
* use fixtures marked as `autouse`

**What does not work**

* parametrized fixtures
* using non-string fixture values in responses (example in repository fails because of this, see #20)

**This needs to be tested thoroughly, I am no pytest expert and do not know if this works in all cases or breaks in other than basic ones.**

Any thoughts/feedback on this would be appreciated.

### DIFF
--- a/example/pytest/conftest.py
+++ b/example/pytest/conftest.py
@@ -1,0 +1,11 @@
+import pytest
+
+
+@pytest.fixture
+def number():
+    return 123
+
+
+@pytest.fixture
+def double(number):
+    return number * 2
--- a/example/pytest/server.py
+++ b/example/pytest/server.py
@@ -1,0 +1,21 @@
+from flask import Flask, jsonify, request
+
+
+app = Flask(__name__)
+
+
+@app.route("/double", methods=["POST"])
+def double_number():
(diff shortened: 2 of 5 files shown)


michaelboulton:
Now that we have markers for tests it seems like it should be fairly easy to just expose `usefixtures` as a marker rather than have a separate block. I'm looking at this in a branch now.

PR #65, opened 2018-03-16
Allows extensions .yml and .yaml  as both valid extensions for YAML

mariocesar:
This allows to have .tavern.yml or .tavern.yaml files, .yml is a valid YAML extension

Before this the pytest file collector just matched .tavern.yaml files as valid tavern files.


michaelboulton:
Thanks for that, never really considered it while writing the discovery function

PR #72, opened 2018-03-24
Specifying an encoding parameter when opening a file

zhaogp:
When i use tavern on windows, there was UnicodeDecodeError because my yaml file contains Chinese Character.

I find that the stream object was encoding with "cp936", which was the default encoding on my windows. Decoding with "utf-8"  is better.

Best regards!


benhowes:
Thanks @zhaogp! We'll add an integration test to this before merging.

Issue #73, opened 2018-03-29
environment variables not loading in included files


justinfay:
I want to use a common file for storing the hostnames of services I am testing, I also want to set the hostnames as environment variables.

```
# file common.yaml
name: Tavern test config
description: hostnames and auth keys for tested services
variables:
  app1_host: "{tavern.env_vars.APP1_HOST}"
```
I am running my tests using pytest with this as the default config
```
APP1_HOST=app1.local py.test integration_tests/ \
    --tavern-global-config integration_tests/common.yaml
```
However it seems like the environment variables aren't being substituted into the config as all my tests fail with errors similar to
```
requests.exceptions.ConnectionError: HTTPConnectionPool
(host='%7btavern.env_vars.app1_host%7d', port=80): 
Max retries exceeded with url: /v2.0/company/
(Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f02660a2898>: 
Failed to establish a new connection: [Errno -2] Name or service not known',))
```

justinfay:
I am using python3.6 with the following versions of pytest and tavern
```
pytest==3.5.0
tavern==0.7.7
```

benhowes:
Closed by #77

michaelboulton:
Fixed in 0.8.1

justinfay:
@michaelboulton thanks for the quick fix and for open sourcing this project.

mirrorkeydev:
Experiencing this issue in Python 3.8.5 and Tavern 1.11.1

This configuration:
```yaml
# file globals.yaml
name: globals
description: Some description
variables:
  host: "{tavern.env_vars.TEST_HOST}"
```
with this command:
```bash
source .env
echo $TEST_HOST # https://localhost:5000
py.test test/ -p no:warnings
echo $TEST_HOST # https://localhost:5000
```
Gives this error:

```
E           requests.exceptions.MissingSchema: Invalid URL '{tavern.env_vars.TEST_HOST}/v1/endpoint': No schema supplied. Perhaps you meant http://{tavern.env_vars.TEST_HOST}/v1/endpoint?

/usr/lib/python3/dist-packages/requests/models.py:387: MissingSchema
```

Simply using the exact same reference`"{tavern.env_vars.TEST_HOST}"` in the test itself runs fine, and using a hardcoded host in the `globals.yaml` file works as well, its once I put the refence in `globals.yaml` that it doesn't substitute.

[...]

PR #77, opened 2018-04-03
Format environment variables into 'included' files so you can specify…

michaelboulton:
… an environmetn varaible in an included file

This was implemented different in an old version of the project - this is how it should have been


michaelboulton:
closes #73

PR #79, opened 2018-04-04
Feature/plugins

### LINKED ISSUE #39: Best practices for flask-testing setup
I currently consider using tavern for testing our backend services and would appreciate some of your thoughts.

Is there any recommended way of setting up a (local) testing environment with flask? Flask itself [provides a test client](http://flask.pocoo.org/docs/0.12/testing/) which can be used in unit tests.

Since spinning up a real test server running on localhost brings other issues with it (clean shutdown, port may already be used etc.), I would like to fall back to this test client.

As far as I can tell, tavern currently only supports requests using the `requests` (+ mqtt, which is not relevant to flask) module. I suppose one could _statically_ add flask test-client support in the `taverns.plugins` module, but this seems a little bit hacky.

- Does anybody have experience in using tavern with a local flask setup, preferrably without spinning up a local server?
- Are there any plans and/or development ongoing regarding the plugin system?
- If no, is there any set of requirements or specification regarding a plugin system? If we decide to use tavern ourselves, contributing here might be of interest.

### LINKED ISSUE #41: proper plugin system
There isn't going to be some kind of huge api for this in a first run through. This is already implemented (though hardcoded for 2 request types) in plugins.py. Assuming plugins are like pytest where you define a file called something like 'tavern_plugins.py' (loaded via setuptools entry point automatically). We do need to think of a good way for plugins to register themselves (possibly just a global variable in the file), and a particular setuptools entrypoint name (eg `tavernplug = myextension.tavern_plugin`)

file has to define 4 functions (run in this order):

- `get_session(**kwargs)` - takes any generic keyword arguments relevant to initialise the 'session'. For example, mqtt need mqtt host, port, authorization, tls settings, etc. etc. This will basically just be passed through as `get_session(**test_spec[session_name])`, where 'session_name' is a block that isn't one of the predefined ones (stages, test_name, includes). Has to return something that can be used as a context manager which will be used throughout the test.

- `get_request_type(session, request_spec, test_block_variables)` - get object with a `run()` method that actually runs the test and returns some kind of object corresponding to a response. `session`, is the session returned above (now inside context manager), `request_spec` is any of the dictionary sub-items for the request (eg for `request` this is url, headers, params. for `mqtt_publish` this is topic, payload, json.). `test_block_variables` is all the variables for using/formatting for that test.

- `get_expected(response_spec, test_block_variables)` - Get the expected result as a dictionary (in whatever format is required for the verifier). `response_spec` is any of the dictionary sub-items for the response (eg for `response` this is status_code, params, etc. for `mqtt_response` this is topic, payload, json).

- `get_verifier(session, expected_response, test_block_variables)` - get a object with a `verify(response)` method that takes the actual response and verifies it against the expected response. `session`, is the session returned above (now inside context manager). `expected_response` is the returned value from get_expected **for that particular response type** - there can be multiple responses per test, so there has to be one verifier for each response. `test_block_variables` is all the variables for using/formatting for that test.

Other things to do:

- update docs to emphasise that there is only 1 request per test stage, but 0 or more responses (for example, MQTT messages don't always expected a response, but HTTP requests always do - need some way to expose this in the plugins like `require_response_block = True`)
- Move existing request type (Requests/MQTT) into another subfolder (eg `contrib/`) and always have them available by default
- Optionally add another function like `validate_schema` which takes the request/response block and validates it. This can just be done in the `get_request_type` entrypoint so it might be redundant.
- Move 'MQTT' block (and any other plugin blocks) in test into another key like `plugins`, eg:

```yaml
test_name: Do a thing

plugins:
  mqtt:
    connect:
       host: example.com

stages:
  - name: do something
    mqtt_publish:
      ...
```

I think doing this and making it work with requests and mqtt will be fairly short work, just shuffling the files around a bit.

michaelboulton:
- (indirectly) closes #39 
- closes #41

Tested with https://github.com/taverntesting/tavern-flask

Works as defined in https://github.com/taverntesting/tavern/issues/41#issue-300646421

The only drawback at the moment is that using it with something like tavern-flask requires a 'flask' block in the top level to specify the connection options, but then if you use 'requests' it will throw an error because it doesn't recognise the 'flask' block. I think fixing this would probably require changing the schema of the tests a bit so I'm hesitant to do that


benhowes:
I've just hacked up a very basic implementation of a plugin which uses pycurl as the requester rather than requests (for gathering more performance info) here https://github.com/taverntesting/tavern-curl.

Observations/Questions:
1) We need a tavern response which is not the Requests response
2) Where should performance data be stored?
3) Where should other arbitrary data be stored in a 'response'?

michaelboulton:
I think performance data is a separate issue, but we will probably just have a (pluggable) function where it takes a dictionary and just dumps it in whatever format at the end of each test (or test run).

Letting each test define which backend would require a fairly major refactor - what do you see as the advantage to being able to do that?

For the response, I guess we just need:
- raw data
- data as json
- headers
- status code

Issue #25, opened 2018-02-06
How to upload images in my request body just like postman?


sridharaiyer:
I want to upload images just like I do it in Postman. See the screenshot of my Postman session.
![screen shot 2018-02-06 at 4 01 39 pm](https://user-images.githubusercontent.com/5684178/35886618-4dabc7fc-0b57-11e8-9c55-992b5e45a247.png)

michaelboulton:
This isn't currently supported at the moment because it's a bit irritating to open all the files that are specified in the input so that requests can pass them. It might be supported in a future version but I don't have any estimate on when it would be finished.

michaelboulton:
This should be addressed in 0.9.0.

PR #82, opened 2018-04-07
Adds file upload functionality

aweidner:
Addresses #25 

The gist of this implementation is that files are opened when the arguments to `requests` are parsed.  A context manager is added to each open file before the request gets submitted so the files will be closed after the request completes.  All context managers are bundled in an `ExitStack` coming from the `contextlib2` module so it should be python 2.7 compatible.

https://contextlib2.readthedocs.io/en/stable/#contextlib2.ExitStack

An example of how this looks added to the YAML for a request

    stages:
      name: Upload multiple files
      request:
      url: "{host}/fake_upload_file"
      method: POST
      files:
        test_files: "test_files.tavern.yaml"
        common: "common.yaml"
      response:
        status_code: 200

I added in an integration test that accepts file uploads and 200s, otherwise if no files were specified it 401s.  If there's a better way to automatically test this functionality let me know and I'd be happy to pursue it.

### DIFF
--- a/tavern/request/rest.py
+++ b/tavern/request/rest.py
@@ -1,4 +1,3 @@
-import functools
 import json
 import logging
 import warnings
@@ -8,6 +7,7 @@ try:
 except ImportError:
     from urllib import quote_plus
 
+from contextlib2 import ExitStack
 from future.utils import raise_from
 import requests
 from box import Box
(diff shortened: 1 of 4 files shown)


michaelboulton:
Thanks for this, it looks good. The only thing I would suggest changing is instead of opening the files and using the open file in a context manager separately, use https://docs.python.org/2/library/contextlib.html#contextlib.closing so the opening/closing can be done in one place, something along the lines of:

```python
from contextlib import closing
...
with ExitStack() as stack:
    for filename in request_args["files"]:
        stack.enter_context(closing(open(filename, "rb")))
```

Thanks for fixing that typo as well

michaelboulton:
Thanks for that, I'll put a note in the docs about it.

benhowes:
Thanks @aweidner!

PR #102, opened 2018-05-15
Add cast for !bool

dbalagansky:


michaelboulton:
Oops, only just noticed this now. I'll ad your test into master as well, thanks!

PR #105, opened 2018-05-25
Add skip/only keywords for selective/isolated stages

cetanu:
Adding this for the following use case:

* During development, one might want to only run a single stage in a test without running everything else (which can be time consuming)
* During development, one might want to skip a test that is flaky, or broken due to a known issue, whilst retaining the stage code so that it can be conveniently un-skipped later

The changes were relatively simple, except for one thing - I had to split the run_test function into two to satisfy pylint (it was reporting too-many-statements for the run_test function 52/50). I hope that's okay.

### DIFF
--- a/tavern/core.py
+++ b/tavern/core.py
@@ -78,47 +78,55 @@ def run_test(in_file, test_spec, global_cfg):
 
         # Run tests in a path in order
         for stage in test_spec["stages"]:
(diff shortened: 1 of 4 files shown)


michaelboulton:
Sorry for the late reply on this, it looks good thanks!

Issue #93, opened 2018-04-25
Problem sending file


luishens01:
Hi, 
we are currently learning Tavern for Rest testing and I have a problem creating a test.

This is an example of the working curl that I need to create in Tavern:

`curl -f -v -i -X POST -H "Authorization:Bearer XXX...XXXX" -F package=@/dir1/dir2/MY_FILE http://url`

and this is the way we are trying to recreate it:

```yaml
  _- name: upload_package
    request:
      url: http://url
      method: POST
      headers:
        authorization: "Bearer {test_login_token:s}"
        Content-Type: "multipart/form-data;
      files: 
        package: "/dir1/dir2/MY_FILE"
    response:
      status_code: 201_
```

This is not working for us and have been trying a lot of combinations with the content-types in headers.

Could you guide us in how we can achieve this ?

Thanks in advance
Luis

michaelboulton:
When you say it's not working, do you mean that it's not sending the file with the correct content type or Tavern is giving an error of some sort? Could give you the output so we can look at it?

Daishi1223:
When we set header `Authorization`, received header `Content-Type` will be `application/json`.

tjacoby:
I have the same problem. For file uploads the `Content-Type` header has to include the boundary information (see https://tools.ietf.org/html/rfc1867#section-6). As @Daishi1223 mentioned the Content-Type header is overridden with `application/json` whenever additional headers are passed.

ChristianSauer:
I have the same problem and it's a showstopper

michaelboulton:
This should be fixed in 0.9.8 whenever Travis deploys it

michaelboulton:
Closing this as it should now be fixed.

PR #108, opened 2018-05-26
Fix for #93

ChristianSauer:
The application-json content type was set even when files where present and a multipart-formdata datatype would be needed.
See https://github.com/taverntesting/tavern/issues/93 for details.
I have fixed the bug and added a warning when somebody tries to set a content-type in that situation since requests does not support that.

### DIFF
--- a/tavern/request/rest.py
+++ b/tavern/request/rest.py
@@ -39,8 +39,6 @@ def get_request_args(rspec, test_block_config):
         BadSchemaError: Tried to pass a body in a GET request
     """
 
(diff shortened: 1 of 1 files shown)


michaelboulton:
As I mentioned in the issue, this should have been fixed in Tavern 0.9.8 (with https://github.com/taverntesting/tavern/commit/df3ce0bbbfa8f74549175a53860c7e009c0066f4). Is it still behaving incorrectly in that release?

michaelboulton:
I'm assuming this is already fixed then.

PR #109, opened 2018-05-26
Allow test variables to be nested

ibaniski:
Update the format_keys() function to change the variables dict to a Box
object. This allows the the test variables to be nested. For example:

  variables:
    users:
      foo:
        username: foo
        password: foo_pass
      bar:
        username: bar
        password: bar_pass

Then in the test spec they can be referenced as "{users.foo.username:s}"
and "{users.bar.username}".

Updated the unit tests "includes" fixture to have nested variables for
the request, and the "req" fixure for the requests test to reference the
nested variables.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -141,19 +141,21 @@ to run the integration tests, make sure you have
 ``tox -c tox-integraton.ini`` (bear in mind this might take a while.)
 It's that simple!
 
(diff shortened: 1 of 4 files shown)


michaelboulton:
Looks good to me. The only thing I would suggest is to add (or, modify I suppose) an integration test to make sure it still works when loading directly from yaml.

Issue #114, opened 2018-05-31
Add support for before/after test stage actions


Birne94:
follow up to this comment: https://github.com/taverntesting/tavern/issues/41#issuecomment-370462578

Currently, the support for pre- and post-stage actions is rather limited. The only ones we support right now are delays and clearing cookies. I suggest adding the possibility to call arbitrary functions here. The format could be similar to the `$ext` function calls currently allowed for validation.

One use case we have for this would be to use [freezegun](https://github.com/spulec/freezegun) to simulate changes in time for a single request (e.g. for testing session expiry).

```yaml
stages:
  - name: test-datetime-endpoint

    before:
      - function: my.package.utils.freeze
        extra_kwargs:
          time: "2018-05-31 12:00:00"

    after:
      - function: my.package.utils.unfreeze

    request:
      url: "{host}/get-time"
      method: GET

    response:
      status_code: 200
      body:
        current_time: "2018-05-31 12:00:00"
```

```python
# my.package.utils

def freeze(time, context):
    t = context['freezegun'] = freezegun.freeze_time(time)
    t.start()


def unfreeze(context):
    t = context.pop('freezegun')
    t.stop()

```

Note the `context` parameter I have added here which allows passing a context object (might just be a dictionary) to the function. This allows a pair of setup- and teardown methods to keep track of created objects. The context may also contain additional info, for example the session is required for clearing the cookies.

The context variable should be injected dynamically, which is controlled either through an attribute in the spec or through a decorator:

```yaml
      - function: my.package.utils.freeze
        with_context: true
        extra_kwargs:
          time: "2018-05-31 12:00:00"
```

```python
# sets freeze._context = True
@tavern.utils.inject_context
def freeze(time, context):
    t = context['freezegun'] = freezegun.freeze_time(time)
    t.start()
```

This allows us to still use functions without context which is useful when using builtins (e.g. `time.sleep` as a replacement for `delay_after` and `delay_before`).

What are your thoughts on this?

michaelboulton:
I have a proof of concept that will handle this kind of thing that will be going into a 2.x release https://github.com/taverntesting/tavern/pull/802 If anyone following this issue wants to give any feedback or examples of what they would actually use this for then I can improve it / add tests for those scenarios.

PR #115, opened 2018-06-04
Add custom before and after handlers for test stages

Birne94:
This PR adds support for custom before- and after handlers as proposed in #114 

An example which uses these handlers to measure response times has been added as well.

```
$ cd examples/timing
$ docker-compose build
$ docker-compose up -d
$ pytest -s
[...]
INFO:utils:request took 0.017 seconds
[...]
```

Currently `delay_before` and `delay_after` options still exist for compatibility reasons, one may either deprecate them for future versions or keep them for simplicity.

### DIFF
--- a/example/timing/Dockerfile
+++ b/example/timing/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3.5-alpine
+
+RUN pip install flask pyjwt
+
+COPY server.py /
+
+ENV FLASK_APP=/server.py
+
+CMD ["flask", "run", "--host=0.0.0.0"]
--- a/example/timing/common.yaml
+++ b/example/timing/common.yaml
@@ -1,0 +1,6 @@
+---
+name: test includes
+description: used for testing against local server
+
+variables:
+  host: http://localhost:5001
--- a/example/timing/docker-compose.yaml
+++ b/example/timing/docker-compose.yaml
@@ -1,0 +1,10 @@
+---
+version: '2'
+
+services:
+  server:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - "5001:5000"
--- a/example/timing/readme.md
+++ b/example/timing/readme.md
@@ -1,0 +1,6 @@
+# Timing example
+
+This example uses custom before and after handlers to measure request
+times.
+
+Run pytest with `-s` or `--capture=no` to see logging output.
--- a/example/timing/server.py
+++ b/example/timing/server.py
@@ -1,0 +1,21 @@
+from flask import Flask, jsonify, request
+
+
+app = Flask(__name__)
+
+
+@app.route("/double", methods=["POST"])
+def double_number():
+    r = request.get_json()
+
+    try:
+        number = r["number"]
+    except (KeyError, TypeError):
+        return jsonify({"error": "no number passed"}), 400
+
+    try:
+        double = int(number)*2
+    except ValueError:
+        return jsonify({"error": "a number was not passed"}), 400
+
+    return jsonify({"double": double}), 200
--- a/example/timing/test_server.tavern.yaml
+++ b/example/timing/test_server.tavern.yaml
@@ -1,0 +1,21 @@
+---
+
+test_name: Make sure server doubles number properly
+
+stages:
+  - name: Make sure number is returned correctly
+    before:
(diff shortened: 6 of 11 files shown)


michaelboulton:
Just to be clear, I have seen this but if we're going to make breaking schema changes with the 1.0 release then I'm hesitant to add this when it's going to be one of the things that will be removed. 'before'/'after' actions are probably going to be offloaded to some hooks in a similar way to this (or fixtures)

Birne94:
e61d680be1a8e31645d41d0082dcda925f84559e should allow doing so by using the context object:

```python
@inject_context
def load_number(context):
    context['variables']['input'] = 5
    context['variables']['output'] = 10
```

michaelboulton:
Following your suggestion in https://github.com/taverntesting/tavern/pull/143#issuecomment-402157560 I'm wondering whether it's better to make tavern-specific stage fixtures and give it the test block config as an argument. These would behave _sort of_ like pytest fixtures but would not be interchangeable and would only affect one stage at a time - I don't want to go through the hassle of re-implementing pytest fixtures to work on a per-stage basis. Pseudocode to handle it:

```python

# In core.py
yield_fixtures = []

for fixture in stage.get("tavern_fixtures", []):
    return_val = fixture(test_block_config)

    if isinstance(return_val, generator): # or however you check this
        yield_fixtures += return_val

response = run_stage(...)

for y in yield_fixtures:
    y.send(test_block_config, response)
```
```python
# in conftest.py?
@tavern.fixture
def time_stage(test_block_config):
    t1 = time.time()

    yield

    t2 = time.time()

    print("Took {}".format(t2 - t1))


@tavern.fixture
def save_response_to_disk(test_block_config):
    response = (yield)

    with open("thing.txt", "r") as wfile:
        wfile.write(response.data)
```
This could then also be used to remove the `$ext` functionality (which is ugly)

Birne94:
My suggestion in https://github.com/taverntesting/tavern/pull/143#issuecomment-402157560 was more along the lines of having a fixture named `tavern` which can be injected into other fixtures as needed. I am not sure, however, how to implement such a fixture and inject the session data properly.

```python
# global fixture file
@pytest.fixture
def tavern():
    proxy = TavernSession(...)
    yield proxy

# conftest.py
@pytest.fixture
def myfixture(tavern):
    tavern.set_variable(foo='bar')
```

Currently, I see little advantage in reinventing the pytest's fixture system. This would have been a good approach if tavern wants to be runnable without pytest, but in the current state we are always running in pytest's context.

When implementing custom fixtures, we need to keep a lot of aspects in mind, which might not be trivial (I think):
- fixture discovery (`conftest.py` may be located anywhere in the directory structure between root and the test file, some might even override each other)
- fixture dependencies
- fixture scopes (session/module/test, see my comment in https://github.com/taverntesting/tavern/pull/143#issuecomment-402165661)

Using pytest fixtures also allows us to reuse setup code between unit and integration tests, e.g. monkey patching, configuration, loading dummy data.

The only issue pytest's fixture system brings is the inability to access the response. Maybe one can find a workaround here, e.g. a custom decorator which sends the response to the fixture iterator:

```python
def tavern_fixture(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        it = func(*args, **kwargs)
        yield next(it)
        it.send(...)
        next(it)
    return wrapper
```

michaelboulton:
The issue is that Pytest fixtures are hardcoded to work on a per-test basis and do not map at all to the 'stages'. They're fine if you just want to set up some initial data, but passing data between stages or accessing the response for a stage using Pytest fixtures isn't possible, as you've noted. I think it would be far more complicated to try and get fixtures to work with stages, and then potentially have to reimplement it when Pytest 4.0 comes out.

My idea for the 'tavern fixtures' would be a much more simple version that only took the test block config as an argument and would only ever be sent the response - no dependencies, no scope, etc.

benhowes:
Hi @Birne94, really sorry we've been quiet on this for so long. @michaelboulton and I are working on something for hooks this weekend. We'll roll your thoughts into whatever we do and thank you so much for the input. It's probably best to just close this one since it's gone stale though. Thanks again!

xu4wang:
Sorry what's the status for this issue, how shall we add a setUp and tearDown function into a tavern test case?

THanks..

Issue #119, opened 2018-06-11
Tavern '1.0' changes


michaelboulton:
There are a few things that need changing and a few issues open that are hard to fix without breaking backwards compatibility. To make sure we know what is going to break in advance, I'm going to try and lay it all out here. This is almost all just related to tighter integration with pytest.

### pytest integration

- `tavern-ci` will either be removed or just become an alias for `py.test`
- A 'base' REST/MQTT request/response object to make it easier for plugins to verify responses

- `delay_after` and `delay_before` will become built-in fixtures or hooks. This is a breaking API change.
- parametrization with pytest - running the same test, but with a different formatted value (or possibly a different dict/list/value)
- marking tests - this will just be for use with the `-m` flag. A list of tags applied to each test which can be used to run a certain subset of tests.
- This is sort of covered by parametrization and fixtures, but we want to be able to add 'plugin' blocks depending on a fixture. eg https://github.com/taverntesting/tavern/issues/113 . This would probably be a breaking API change, but I doubt many people are using plugins at the moment so it shouldn't be too bad.

Either/both:
- Some sort of integration with pytest fixtures - unsure how this will work now, but it will allow you to do some arbitrary python code and return a dictionary/list/value for the request/response/etc. This will probably be implemented such that we can `send` a value to the fixture if it's a generator, so fixtures can be written like this:
```python
@pytest.fixture(name="tavern_complicated_request_val")
def test_complicated_return_value():
    val = calculate_something_complicated()

    response = (yield val)

    expected = calculate_complicated_response()
    assert response.body["value"] == expected    
```
```yaml
request:
  json:
    sent_value: !fromfixture "tavern_complicated_request_val"
```
- extra hooks to allow people to run things before/after/during tests. For example, a hook called `tavern_before_test` which is run before every test and is called with the associated `request` and `YamlItem`.

Either one of these should cover all use cases, but both of them being available would definitely be the best case scenario.

### Possible

- xfailing tests - this is probably going to be hidden from external use but it would be helpful to be able to write a deliberately broken test for tavern to make sure that it can be caught and reported to the user appropriately.
- skipif with pytest - eg, skip a test if running against a certain server. This will provide similar but different functionality to marking tests. I'm not sure of a nice way to implement this other than using `exec()`, but that might just have to happen.

### problems

[...]

PR #121, opened 2018-06-13
Add 'strict' argument to recursive key checking so dictionary key mis…

michaelboulton:
…matches can be ignored or made more strict

This adds a 'strict' argument which can be used from the config file, command line, in a test, or in a 'response' block for a test stage.

If not set, this will use the 'legacy' Tavern behaviour where top level keys are ignored but everything below is matched strictly. If set to False, it will ignore extra keys present in the response at all levels. If set to True, it will fail tests if the response does not exactly match the expected response.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -91,7 +91,9 @@ addopts =
     --cov-report term-missing
     --cov tavern
     --doctest-modules
-    -r xs -v --strict
+    -r xs
+    -vv
+    --strict
     -p no:logging
 norecursedirs =
     .git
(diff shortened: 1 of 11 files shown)


benhowes:
What's the use-case for being able to toggle this on and off? Are there situations where you may want it turned off on a dev instance, but on when in integration? 

I can see the value in having it in the tests, but I don't get it as a global at the moment.

michaelboulton:
It's a global at the moment mainly to maintain backwards compatability #119

PR #122, opened 2018-06-13
Feature/mqtt qos

### LINKED ISSUE #104: Subscription qos for MQTT
Should be able to set a QoS level for subscribing to a topic. Easy to do but not doing until #79  is merged

michaelboulton:
Add choice of mqtt qos when expecting a response

closes #104


benhowes:
Merging because it looks like the docs are up to date too

PR #124, opened 2018-06-14
Feature/gitlab ci

michaelboulton:


michaelboulton:
Although this would be better than travis, there are a few issues:

1. Gitlab CI can take a long time to start running and this also interacts badly with the Github PR interface (see https://gitlab.com/gitlab-org/gitlab-ce/issues/45950)
2. Running docker builds is more complicated because of the way it works in Gitlab CI

PR #125, opened 2018-06-15
WIP: Ability to load external test spec components (stages)

ibaniski:
Extend the test schema to permit loading external stages into any test
spec. Do this by extending the YAML parser to include two new
constructors, !load and !stage.

!load <external_file.yaml>
---------------------------
Loads test spec stages from an external YAML. The stages are just made
available to the test spec; they are not executed at this point. The
format for the file is similar to the "stages" subdocument:
```
stages:
    stage_x:
      - name: Stage X
      - request: ...
```

!stage <stage_id>
-----------------
Extend the test spec with the stage. The YAML document is actually
extended, as if the stage were written at this spot. This means things
like anchors, saved elements, etc. are all available. Example usage
would be:
```
  test_name: Test ABC
  loads:
    - !load <external_file.yaml>
  stages:
    - !stage stage_x
```

Why do this? Anchors partially address this in the same file. This can
be a little incovenient for stages that are needed in all tests (i.e.
tests in different files). An example would be a stage that
authenticates and stores a token that will be used in subsequent stages.

The goal of this implementation is to be extensible. For example, one
may want to define a complete response or request headers externally
that can be reused across tests.

Testing
-------
New integration test added for external components (stages). It defines
an authentication stage that is then used in 2 separate tavern tests.

### DIFF
--- a/tavern/util/loader.py
+++ b/tavern/util/loader.py
@@ -75,7 +75,7 @@ class IncludeLoader(Reader, Scanner, Parser, RememberComposer, SafeConstructor, 
     def get_data(self):
         self.loaded_stages = {}
 
-        return super(self.__class__, self).get_data()
+        return super(IncludeLoader, self).get_data()
 
 
 def _get_yaml_filename(loader, node):


Birne94:
For what its worth, you can also use the `!include` tag for this. We have been using this feature to include common stages (e.g. login/authentication) for our tests for a while.

Example:

```
# test.tavern.yaml
stages:
  - !include common/login.yaml
```

```
# common/login.yaml
name: login
request:
  method: POST
  url: "{host}/auth"
  headers:
  json:
    method: signed_token
    token: "{token:s}"
response:
  status_code: 200
  body:
    access_token: !anystr
    refresh_token: !anystr
    $ext:
      function: tavern.testutils.helpers:validate_jwt
      extra_kwargs:
        jwt_key: access_token
        key: test-jwt-secret
        options:
          verify_signature: true
  save:
    body:
      access_token: access_token
      refresh_token: refresh_token

```

ibaniski:
@Birne94 you are right, maybe I was a little too eager here. I did try to make ```!include``` work, but gave up. This should resolve my immediate road block with reusing stages.

However, at least from the documentation, it feels like this is overloading/side-effect of the ```!include``` operator. I combed through the documentation several times, and all I could get for using ```!include``` was about controlling variables. The code suggested that the main purpose is to load the variables and use them later when the tests are compiled (rather than manipulate the YAML document) -- I admit this is me reading into it, hopefully this is where the maintainer (@michaelboulton) can comment on intended usage.

Without complete understanding of the architecture + future, here's why I think this approach is good:
* Can formally define the stages + explicitly use them in the test spec
* Can group multiple related stages in a single YAML
* Can be extended to have other test subcomponents (I guess you can do this with ```!include``` now)

I will fall back to using ```!include``` until I hear from @michaelboulton whether this aligns with what's planned for this project or not. Thanks @Birne94!

michaelboulton:
That `!include` behaviour is not all all what I was expecting anyone to use it for, but if it works then it works I suppose. I'll look at this now

tavern/util/loader.py
```diff
@@ -82,13 +90,54 @@ def construct_include(loader, node):
...
+    return None
+
+def construct_stage(loader, node):
+    """ Construct a stage from previously loaded stage.
+    """
+    stage_name = str(loader.construct_scalar(node))
+    try:
+        stage = loader.loaded_stages[stage_name]
+    except KeyError as e:
+        logger.error("The following key does not exist: %s", e.args[0])
```
michaelboulton: Does this depend on the order that things are defined in the file? eg, does it work if you have this:

```yaml
stages:
  - !stage abc

loads:
  - extra_stages.yaml
```

tavern/util/loader.py
```diff
@@ -82,13 +90,54 @@ def construct_include(loader, node):
...
+
+
+def construct_include(loader, node):
+    """Include file referenced at node."""
+
+    filename = _get_yaml_filename(loader, node)
     with open(filename, 'r') as f:
         return yaml.load(f, IncludeLoader)
 
+def construct_load(loader, node):
```
michaelboulton: Is there a reason that you put the `loads` block as a separate thing to `includes`? In 'config' files there is a `variables` block which was there so that it could be extended in future (such as by adding `stages`), is there anything getting in the way of reusing `includes`?

This also has the advantage of being able to move the 'login' stage to a config file which can then be swapped out depending on which server is being tested against. eg:

```yaml
stages:
  - !stage login
  - name: get all users
    ...
```
Then the login stage can automatically be swapped with
```shell
$ pytest --tavern-global-cfg test-local.tavern.yaml
$ pytest --tavern-global-cfg test-integration.tavern.yaml
```

tavern/util/loader.py
ibaniski: Order is definitely important, what you've listed will not work with the current implementation. I don't know how strongly you feel about making order not relevant; in every config I've used (YAML or not) if including of external resources is supported, the resources need to be included before they can be referenced.

tavern/util/loader.py
ibaniski: No. My initial thought was to just reuse ```includes```, but opted for this approach to minimize impact on existing functionality.

However, I think the actual "login stage" should be the same no matter the environment. One can/should still use the variables like in your example to parameterize the different server (or whatever is different between environments).

ibaniski:
@michaelboulton let me know if you are OK with the general approach. I can refactor the code to reuse ```includes```.

michaelboulton:
I do think it's an excellent approach, I just think it's better to keep it in the 'includes' to avoid fragmenting the config files more.

PR #129, opened 2018-06-18
Feature/mark parametrize

michaelboulton:
Parametrizing format values. Because pytest does this internally with funcargnames and fixturenames etc. it's implemented manually here.

- #8 
- #84


benhowes:
This is really cool! What happens if you want to parameterize on more than one value?

michaelboulton:
I've added functionality for parametrizing multiple things, though it is slightly hacky because of Python 2 support

PR #130, opened 2018-06-18
Feature/change tavern ci

michaelboulton:
Make it so tavern-ci just calls pytest

### DIFF
--- a/tavern/core.py
+++ b/tavern/core.py
@@ -19,7 +19,7 @@ from .printer import log_pass, log_fail
 
 from .plugins import get_extra_sessions, get_request_type, get_verifiers, get_expected
 
-from .schemas.files import verify_tests
+from .schemas.files import verify_tests, wrapfile
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


tavern/core.py
```diff
@@ -155,7 +158,32 @@ def run_stage(sessions, stage, tavern_box, test_block_config):
...
+    if kwargs:
+        warnings.warn("Passing extra keyword args to run() when using pytest is used are ignored.", FutureWarning)
+
+    if pytest_args is None:
+        import sys
+        sys.argv += ["-k", in_file]
+    else:
+        pytest_args += ["-k", in_file]
+
+    return pytest.main(args=pytest_args)
```
benhowes: Is it possible to override the test collection in order to reduce the time penalty for this?

benhowes: If we can do that, I suggest that we drop the `run_legacy` straight away because it's another thing to maintain.

michaelboulton: agreed

PR #136, opened 2018-06-21
Fix parametrization without includes

gdyuldin:
Parametrization test without `insudes` raised KeyError


michaelboulton:
Oops, thanks for spotting that!

PR #137, opened 2018-06-22
Fix logging simple response body

gdyuldin:
Is case of response body contains valid json like int or string or None,
logging it raised an Attribute error.


michaelboulton:
Thanks for the fix!

PR #138, opened 2018-06-24
feature/jmes_validations: Added new helper functions

abhimanyupandian:
These helper functions can perform API response validations using JMES path expressions.


setup.py
```diff
@@ -69,9 +69,9 @@ def run(self):
     "mock",
     "paho-mqtt==1.3.1",
     "pytest-remove-stale-bytecode",
+    "jmespath",
```
michaelboulton: This doesn't need to be in here if it's already in the setup.cfg requirements

tox-integration.ini
```diff
@@ -22,6 +22,7 @@ deps =
     paho-mqtt
     mqtt: fluent-logger
     pytest-remove-stale-bytecode
+    jmespath
```
michaelboulton: Same as above, this doesn't need to be here

tavern/testutils/helpers.py
```diff
@@ -1,16 +1,59 @@
...
+    'none': type(None),
+    'number': (int, float),
+    'int': (int),
+    'float': float,
+    'bool': bool,
+    'str': str,
+    'list': list,
+    'dict': dict
+}
+
```
michaelboulton: Just a minor thing - could you move the extra functions/definitions (eg, `test_type`, `COMPARISONS`, `regex_compare`, etc) to a file `tavern/testutils/jmes.py` to keep it encapsulated?

example/simple/test_comparisions.tavern.yaml
```diff
@@ -0,0 +1,19 @@
...
+    response:
+      status_code: 200
+      body:
+        id: 1
+        $ext:
+          function: helper:validate_content
+          extra_kwargs:
+            comparisions:
+              - {jmespath: "userId", operator: "eq", expected: 1}
+              - {jmespath: "userId", operator: "type", expected: str}
```
michaelboulton: Could you add a few more integration tests against the local test server as well? The `/fake_dictionary` endpoint returns a dictionary like this:

```json
{
        "top": {
            "Thing": "value",
            "nested": {
                "doubly": {
                    "inner": "value",
                }
            }
        },
        "an_integer": 123,
        "a_string": "abc",
        "a_bool": true
    }
```
so you could check that accessing nested keys works as well.

PR #140, opened 2018-06-27
Add support of many keys in parametrization

gdyuldin:
Py.test allows to use many parameters in a parametrize. Like

```python
@pytest.mark.parametrize(['foo', 'bar'], [(1, 2), (3, 4)])
def test_baz(foo, bar):
   ....
```

This patch add supporting of the same functionality with next syntax

```yaml
marks:
  - paramterize:
      keys:
       - foo
       - bar
     vals:
       - foo: 1
         bar: 2
       - foo: 3
         bar: 4
```


michaelboulton:
I think this case should already be covered - example from the [documentation](https://taverntesting.github.io/documentation#parametrize):

```yaml
marks:
  - parametrize:
      key: fruit
      vals:
        - apple
        - orange
        - pear
  - parametrize:
      key: edible
      vals:
        - rotten
        - fresh
        - unripe
```
So your example could be done like this:

```yaml
marks:
  - paramterize:
      key: foo
      vals:
       - 1
       - 3
  - paramterize:
      key: bar
      vals:
       - 2
       - 4
```

This lets you parametrize multiple values (with a different number of values for each thing you're parametrizing).

Is there something different you need that can't be done with the existing implementation?

gdyuldin:
Let's suppose we have a rest api, that returns reversed input argument. Something like:
`{'a': 'foobar'} -> {'a': 'raboof'}`

It might be tested with a pairs of arsument (`to_send`, `expected`)
In case of using many `parametrize` test:
```yaml
marks:
  - paramterize:
      key: to_send
      vals:
       - foo
       - bar
  - paramterize:
      key: expected
      vals:
       - oof
       - rab
```
it will produce next pairs
```yaml
to_send: foo
expected: oof
---
to_send: foo
expected: rab
---
to_send: bar
expected: oof
---
to_send: bar
expected: rab
```

But we need only
```yaml
to_send: foo
expected: oof
---
to_send: bar
expected: rab
```

tests/integration/test_parametrize.tavern.yaml
```diff
@@ -219,3 +219,31 @@ stages:
       status_code: 200
       body:
         value: "{to_send}"
+
+
+---
```
michaelboulton: I think this could just do with 1 extra test to make sure that it can handle both `key` and `keys` in one test - eg:

```yaml
marks:
  - parametrize:
      key: abc
      vals:
        - one
        - two
  - parametrize:
      keys:
        - def
        - ghi
      vals:
        - def: three
          ghi: four
        - def: five
          ghi: six
```

tavern/testutils/pytesthook.py
```diff
@@ -150,16 +170,16 @@ def get_parametrized_items(self, test_spec, parametrize_marks, pytest_marks):
             # 'raise x from y' syntax for exception causes. This will still
             # raise an error at validation, but the test name will be a bit
             # mangled
-            vals = [["TAVERNERR"]]
+            # TODO: fix
```
michaelboulton: Is this TODO meant to be in here?

tavern/schemas/extensions.py
```diff
@@ -13,6 +14,23 @@
...
+    @functools.wraps(func)
+    def wrapper(value, rule_obj, path):
+        schema = rule_obj.schema
+        if schema is not None:
+            schema = schema.copy()
+            del schema['func']
+            c = core.Core(source_data=value, schema_data=schema)
+            c.validate(raise_exception=True)
+        else:
+            print(value, rule_obj, path)
```
michaelboulton: Is this print here for a reason? It should be removed, raise an exception, or changed to a log message if it is useful in some way

tavern/schemas/extensions.py
```diff
@@ -13,6 +14,23 @@
...
+    """Decorator to execute custom validator after all nested validation will
+    be passed."""
+    @functools.wraps(func)
+    def wrapper(value, rule_obj, path):
+        schema = rule_obj.schema
+        if schema is not None:
+            schema = schema.copy()
+            del schema['func']
+            c = core.Core(source_data=value, schema_data=schema)
+            c.validate(raise_exception=True)
```
michaelboulton: Can you add some documentation to what this is doing? I get that it does some extra validation on the input but I don't understand why or how

tavern/testutils/pytesthook.py
```diff
@@ -107,6 +108,12 @@ def pytest_addoption(parser):
     )
 
 
+def merge_dicts(x, y):
```
michaelboulton: Should use `tavern.util.dict_util.deep_dict_merge` instead of defining this function which does the same thing

michaelboulton:
This has now been implemented slightly differently , see documentation https://taverntesting.github.io/documentation.html#parametrize

PR #141, opened 2018-06-28
Support for streaming/downloading files

shasha79:
Added support and test for streaming/downloading files API using stream=True parameter for session.request() method (its usage is described here: http://docs.python-requests.org/en/master/user/advanced/#body-content-workflow)

### DIFF
--- a/tavern/_plugins/rest/request.py
+++ b/tavern/_plugins/rest/request.py
@@ -87,7 +87,7 @@ def get_request_args(rspec, test_block_config):
             rspec["headers"] = {i: j for i, j in headers.items() if i.lower() != "content-type"}
     elif headers:
         # This should only be hit if we aren't sending a file
-        if not has_content_header and not rspec.get("stream"):
+        if not has_content_header:
             rspec["headers"]["content-type"] = "application/json"
 
     fspec = format_keys(rspec, test_block_config["variables"])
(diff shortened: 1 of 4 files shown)


tests/integration/test_stream.tavern.yaml
```diff
@@ -0,0 +1,15 @@
+---
+
+test_name: Test streaming (downloading) file
+
+includes:
+  - !include common.yaml
+
+stages:
+  - name: Stream file
```
michaelboulton: Should this stage have the 'stream' flag?

tavern/_plugins/rest/request.py
```diff
@@ -86,7 +87,7 @@ def get_request_args(rspec, test_block_config):
             rspec["headers"] = {i: j for i, j in headers.items() if i.lower() != "content-type"}
     elif headers:
         # This should only be hit if we aren't sending a file
-        if not has_content_header:
+        if not has_content_header and not rspec.get("stream"):
```
michaelboulton: I don't understand why this check is here - the streaming should affect the response, not the request content-type header

michaelboulton:
Thanks for that, It should be deployed in version 0.14.5

shasha79:
Thanks for the great library, happy to contribute!

Issue #113, opened 2018-05-31
Plugin config through global config


Birne94:
When using a plugin that requires additional configuration (e.g. `tavern-flask`), the config needs to be provided in every test file at least once. Multiple tests within one file may reference it through anchors.

Example:

```yaml
flask:
  app:
    location: tests.integration.server:app
```

Since most likely the whole test suite will use the same plugins, it would be a good QoL change to be able to configure plugins through the global config, specified through `tavern-global-cfg` in either the `pytest.ini` or per command line.

In order to keep this as flexible as possible, I would suggest adding a global configuration option whose values are inserted into all test-specs.

Example global-config.yaml
```yaml
inject:
  flask:
    app:
      location: tests.integration.server:app
```

Example test.tavern.yaml

```yaml
test_name: test_api_auth

stages:
 - ...
```

Which results in a test spec equal to

```yaml
test_name: test_api_auth
flask:
  app:
    location: tests.integration.server:app
stages:
 - ...
```

The implementation should be relatively trivial since test spec and global config are both available in `YamlItem::runtest` and we can simply update the spec dictionary here with the additional values. I guess the implementation for `tavern-ci` would be similar.

@michaelboulton would this be a desirable approach? If so I can submit a PR.

PR #143, opened 2018-07-03
Feature/fixtures

michaelboulton:
This just reuses the `usefixtures` mark so you can specify a list of fixture names to use for that test. This can be used with 'yielding' fixtures to run stuff before and after the test stage, and the return value of any fixtures is added to the variables for that test. For example, this fixture:

```python
@pytest.fixture(name="json_val")
def get_json_val():
    return "abc123"
```

will be put into the variables as `{"json_val": "abc123"}` for use in formatting.

### DIFF
--- a/tavern/schemas/extensions.py
+++ b/tavern/schemas/extensions.py
@@ -166,6 +166,7 @@ def validate_status_code_is_int_or_list_of_ints(value, rule_obj, path):
 
 
 def check_usefixtures(value, rule_obj, path):
+    # pylint: disable=unused-argument
     err_msg = "'usefixtures' has to be a list with at least one item"
 
     if not isinstance(value, (list, tuple)):
--- a/tavern/schemas/extensions.py
+++ b/tavern/schemas/extensions.py
@@ -165,6 +165,18 @@ def validate_status_code_is_int_or_list_of_ints(value, rule_obj, path):
     return True
 
 
+def check_usefixtures(value, rule_obj, path):
+    err_msg = "'usefixtures' has to be a list with at least one item"
+
+    if not isinstance(value, (list, tuple)):
+        raise BadSchemaError(err_msg)
+
+    if not value:
+        raise BadSchemaError(err_msg)
+
+    return True
+
+
 def validate_data_key_with_ext_function(value, rule_obj, path):
     """Validate the 'data' key in a http request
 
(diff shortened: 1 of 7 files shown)


michaelboulton:
It might be worth modifying this slightly before merging so that it can add context to the whole test as well. It is technically possible with this PR to write a fixture like this:

```python
@pytest.fixture(name="use_tavern_flask")
def add_flask_spec(request):
    request._pyfuncitem.spec["flask"] = {"app": ....}
```

but maybe exposing that properly and making it the 'preferred' way of doing it would be good and would fix #113. Thoughts @benhowes @Birne94  ?

Birne94:
How do you feel about introducing a special `tavern` fixture which provides access to different helper methods? These could include setting plugin config, global variables etc.

This would give tavern more control over what can be manipulated in contrast to letting the user manipulate the dictionaries at will and maybe causing cryptic/hard to track errors as a result.

Birne94:
Also, while testing I noticed that the fixture scope `module` does not work with tavern currently. `function` and `session` work as expected.

This can be resolved by changing the base class of `YamlFile` (`tavern/testutils/pytesthook.py`) to `pytest.Module`. `Module` itself is subclass of `pytest.File`, so no behavior should change here.

I am not sure if the additional logic in the `Module` class can cause any trouble, though.

michaelboulton:
How would you expect the scopes to work? I wasn't really thinking about it while implementing this, would the 'module' be the whole file?

Birne94:
By default pytest supports the following scopes:

- session
- module
- class
- function

The `class` session would not make too much sense with tavern, I guess, but `module` would be nice to have in order to setup a common ground for related tests and avoid having to allocate resources multiple times. Since tavern allows defining multiple tests in a single file, I would consider this file a module. This also matches the default behavior of each file being a module.

The other ones (`function` and `session`) work through the default implementation.

michaelboulton:
I suppose it probably works when you inherit from `Module` because it keys the tests on the `fspath` rather than `name`? It seems to not raise any errors but it seems like it might cause some weird errors with fixtures seeing as Tavern doesn't really map to the concept of modules or functions

michaelboulton:
It might be worth actually checking so that if fixtures don't have either 'session' or 'function' scope then it raises an error

michaelboulton:
Trying to use a module scoped fixture fails anyway, and class fixtures seem to just be regenerated between tests, so I think that should be fine as it is

PR #150, opened 2018-07-10
Type convert boolean variables correctly (closes #149)

### LINKED ISSUE #149: !bool casts all string values to True
Tavern provides a `!bool` constructor that is used during configuration for [type conversion](https://taverntesting.github.io/documentation#type-conversions).

Unfortunately, it uses Python's underlying [`bool` constructor](https://docs.python.org/3.7/library/functions.html#bool) which casts all non-empty strings to `True`.

```yaml
value: !!bool "True"
# True
value: !bool "{v_true}"
# True

value: !!bool "False"
# False
value: !bool "{v_false}"
# True
```

[This integration test](https://github.com/taverntesting/tavern/blob/61065bd294b03a33caa0d2ef15fc2908181b5964/tests/integration/test_typetokens.tavern.yaml#L201) would fail if the value was checked.

tommilligan:
See #149 for issue details.

This PR:
- adds failing tests that show `!bool` type conversion is incorrect
- implements a fix for `!bool` using a custom `bool` constructor class

The custom bool constructor is similar to the builtin `bool` constructor as per [PEP 285](https://www.python.org/dev/peps/pep-0285/#specification), and as discussed [here](https://stackoverflow.com/a/21732186/9923936)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -12,3 +12,4 @@ ignore-imports=yes
 
 [TYPECHECK]
 ignored-classes=RememberComposer
+ignored-modules=distutils
--- a/tavern/util/loader.py
+++ b/tavern/util/loader.py
@@ -229,10 +229,24 @@ class FloatToken(TypeConvertToken):
     yaml_tag = "!float"
     constructor = float
 
+class BoolLiteralEval(object):
+    """Using `bool` as a constructor will evaluate all strings to `True`.
+    Using `ast.literal_eval` may result in other datatypes being returned.
+    This constructor returns a boolean given a boolean string representation.
+
+    See discussion here: https://stackoverflow.com/a/21732186/9923936
+    """
+    def __new__(cls, s):
+        if s == 'True':
+            return True
+        elif s == 'False':
+            return False
+        else:
+            raise ValueError("Cannot literally eval {} to a bool".format(s))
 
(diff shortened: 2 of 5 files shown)


michaelboulton:
Rather than using a custom constructor it is probably better to use [strtobool](https://docs.python.org/2/distutils/apiref.html#distutils.util.strtobool) from the standard library, which can handle a lot more than just `True` and `False` (including all the weird yaml ones like `on` and `y`)

Other than that, this all looks good!

tommilligan:
Happy to use `strtobool` instead, if it's okay to allow a wider range of values. Agree it's better than a custom implementation.

See the new commit for changes:
- it still has to be wrapped in a constructor `__new__` method to match other constructors
- `strtobool` actually returns `0|1`, so it's wrapped in `bool`

tommilligan:
There is a [long running known issue](https://github.com/PyCQA/pylint/issues/73) with combining `distutils`, `pylint` and `virtualenv` (which `tox` uses internally). `pylint` reports a false positive that `distutils` cannot be found.

I have added `distutils` as an ignored module in `.pylintrc`.

tommilligan:
@michaelboulton PyPI latest version appears to be 0.14.4 - have had a quick look at the Travis logs and can't see any obvious errors

Issue #153, opened 2018-07-11
pykwalify schema considers typetokens invalid


tommilligan:
When trying to use the pykwalify validated schema with type converted variables, a `BadSchemaError` is thrown. For example:
```yaml
variables:
  verify-ssl: False

---

test_name: Test health
stages:
- name: health
  request:
    url: ...
    method: POST
    json:
      ...
    verify:
      value: !bool "{verify-ssl}"
```
will throw something similar to
```
E   pykwalify.errors.SchemaError: <SchemaError: error code 2: Schema validation failed:        
E    - Value '<tavern.util.loader.BoolToken object at 0x7f388526c4a8>' is not of type 'bool'. Path: '/stages/0/request/verify'.: Path: '/'>
```

tommilligan:
Added failing tests in PR #154, will look at possible fixes this week

michaelboulton:
Fixed in #154

PR #154, opened 2018-07-11
Consider TypeTokens valid types when validating schema

tommilligan:
See issue #153 for description

### DIFF
--- a/tavern/core.py
+++ b/tavern/core.py
@@ -179,7 +179,7 @@ def run_test(in_file, test_spec, global_cfg):
                 )
 
             # Wrap run_stage with retry helper
-            run_stage_with_retries = retry(stage)(run_stage)
+            run_stage_with_retries = retry(stage, test_block_config)(run_stage)
 
             try:
                 run_stage_with_retries(sessions, stage, tavern_box, test_block_config)
@@ -210,7 +210,7 @@ def run_stage(sessions, stage, tavern_box, test_block_config):
 
     expected = get_expected(stage, test_block_config, sessions)
 
-    delay(stage, "before")
+    delay(stage, "before", test_block_config["variables"])
 
     logger.info("Running stage : %s", name)
     response = r.run()
@@ -221,7 +221,7 @@ def run_stage(sessions, stage, tavern_box, test_block_config):
         test_block_config["variables"].update(saved)
 
     tavern_box.pop("request_vars")
-    delay(stage, "after")
+    delay(stage, "after", test_block_config["variables"])
 
 
 def _run_pytest(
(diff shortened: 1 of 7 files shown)


tommilligan:
Added three failing tests that show `tavern` variables cannot be used as part of `pykwalify` validated schema. This makes sense, as at validation time these fields will be `Tokens` rather than the primitive types they promise.

The above commits allow all `bool` values in the current test schema to be variables, fixing one failing test. Before I fix the others thought I'd get some feedback.

**Details**
I initially tried using `isinstance` to extend class checks, but this threw up bugs (e.g. pykwalify allows integers to count as floats). So, I've just extended pykwalify's [type check functions](https://github.com/Grokzen/pykwalify/blob/a41db86ebfb9a0ef00d7682efe7a29b76b052289/pykwalify/types.py#L76) to [include `Tokens`](https://github.com/taverntesting/tavern/pull/154/files#diff-d5fe3ada6f9cf0cb9db3de78db3f3fb6R21). It works nicely!

To accept `tavern` variables as part of a schema file, we use `pykwalify` extension syntax:
```yaml
# Allows boolean primitive only
verify:
  type: bool

---

verify: False
```
becomes
```yaml
# Allows boolean primitive or boolean variable
verify:
  type: any
  func: bool_variable

---

verify: "{v_false}"
```

Other schemas will have to opt-in to this approach. If no changes are made to the schema, variables will be invalid.

tommilligan:
@michaelboulton could I get some feedback on whether you'd be happy with the above approach, before I progress it any further?

michaelboulton:
It looks fine to me, I'm just not sure what you mean by:

> Other schemas will have to opt-in to this approach

You mean schema for plugins?

tommilligan:
Yes, exactly. Any schema that uses the pykwalify builtin types `bool`, `int`, `float` etc. would fail validation if given variables. Schema writers would have to explicitly use `bool_variable` etc. to allow variables as well.

ke4roh:
This appears to be the actual issue I remarked about in a recent comment on Issue #149.  Allow me to offer my enthusiastic support for this PR.

benhowes:
@tommilligan I know this one is quite old, but are you in a position that you could take a look so we can get this merged in?

benhowes:
@tommilligan ping!

tommilligan:
@benhowes ready for final review. It's slightly messy as there's not a single point in the code where variables are resolved. I'm happy it will fix _most_ issues, but not certain it will catch _all_ of them.

Essentially:
- when variables are part of the schema, they are loaded as python `Token` instances
- we use functions like `is_float_like` to check them in python
- we use functions like `int_variable` to check them in the pykwalify yaml schema

See above for how this will affect custom schemas.

michaelboulton:
closes #153

michaelboulton:
Thanks a lot for this, this should fix most of the lingering annoying issues with type tokens in various situations.

PR #155, opened 2018-07-12
NTLM Auth Support for Rest

girishrd2004:
Modifications to Rest plugin to support NTLM based authentication

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 pyyaml
 pykwalify>=1.6.1
 requests
+requests_ntlm
 pyjwt
 python-box
 future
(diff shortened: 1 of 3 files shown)


benhowes:
Thanks for the contribution! I'm not sure if @michaelboulton will be okay with this being in the main requests plugin, or if this would be better as a separate plugin (as to keep the main one lighter). Perhaps you can comment @michaelboulton?

Either way, this will need some tests adding please and ideally an example please.

michaelboulton:
I think it should be fine adding it in to the main plugin, there isn't a strict requirement on supporting all possible auth methods in plugins (at least, not yet).

As @benhowes  mentioned though, this does need some tests and/or examples (though not sure how easy that is to do)

benhowes:
We'll close this one off for now due to the lack of tests and it having gone stale. we'd be really happy for you to reopen it with some testing in place so that everyone can benefit from it.

PR #161, opened 2018-07-27
Adding str method to ReprdError class

tejinderss:
With the current tavern-beta-new-traceback implementation, it will not work with junitxml report. In order to fix that I added the str method which pytest reads to output to stdout. This fixes the junit xml report.


tejinderss:
Looking into test failure

michaelboulton:
Thanks for this, I don't really use the junit xml so I never noticed that was broken.

PR #162, opened 2018-07-27
Feature/list strictness change

### LINKED ISSUE #133: Check for one item in a list
If a server returns this:

```yaml
- a
- b
- c
- d
- e
```

and I just want to make sure `e` is there and ignore the rest. This will probably just require expanding what 'strict' key checking can do

michaelboulton:
closes #133


tests/unit/test_utilities.py
```diff
@@ -250,6 +250,21 @@ def test_value_case_matters(self):
         with pytest.raises(exceptions.KeyMismatchError):
             check_keys_match_recursive(a, b, [])
 
+    def test_match_list_items(self):
```
benhowes: Can you test for cases where non-strict matching should fail? perhaps `a=[1,2,3], b=["a"]`

tavern/util/dict_util.py
```diff
@@ -301,12 +301,28 @@ def _format_err(which):
...
+                for i, e_val in enumerate(expected_val):
+                    for a_val in actual_val:
+                        try:
+                            check_keys_match_recursive(e_val, a_val, keys + [i], strict)
+                        except exceptions.KeyMismatchError:
+                            pass
+                        else:
+                            logger.debug("One (or more) list items present - continuing due to strict=%s", strict)
+                            # TODO
+                            # do we need to check more here?
```
benhowes: What were you thinking here?

tavern/util/dict_util.py
```diff
@@ -301,12 +301,28 @@ def _format_err(which):
...
-
             # TODO
             # Check things in the wrong order?
 
+            if not strict:
+                # Iterate over list items to see if any of them match...
+                for i, e_val in enumerate(expected_val):
+                    for a_val in actual_val:
+                        try:
+                            check_keys_match_recursive(e_val, a_val, keys + [i], strict)
```
benhowes: Just a discussion point, but I can see someone raising an issue of wanting to strictly match a single object in a list! If we go down this route, are we going to end up slowly re-implementing pykwalify or JSON schema?

michaelboulton: I'm not sure what you mean by ' strictly match a single object in a list', isn't that what it's doing already?

michaelboulton: Although looking at it again it does need to be changed so that all expected items are present

tavern/util/dict_util.py
michaelboulton: :man_shrugging: I think I wrote that a while ago, it can be removed now I think

michaelboulton:
@benhowes Can you accept it if it's all done?

michaelboulton:
I was just thinking of this and I realised that the 'default' strictness in 1.0 might be a bit annoying with this. example, if the server returns `[1, 2, 3]` and I just want to check that `2` is there I can use the behaviour in this branch. However, if I only expect the server to return `[2]` then that same test will pass but there's extra things being returned that I didn't expect, and unless I looked at the logs I might miss it.
Enabling strict key checking would get around it, but then you need to specify a whole load of extra keys that you might not care about, or have a big test like this:
```yaml
body:
  - !anything
  - !anything
  - 2
  - !anything
...
```

benhowes:
@michaelboulton Are you proposing that we look for a way around the problem of matching a single item in a list? Is there a way we could use a loader for these too? 

I do want to avoid us having an explosion in the number of yaml sentinals, however we are lacking a certain expressiveness. I guess the other alternative is to provide a few builtin external matching functions to ease this kind of situation? Perhaps the latter is a better fit? Either way, ending up with the strict matching thing you mention is not great!

michaelboulton:
@benhowes Along with #273 I suggest that we close this PR and just say that any arbitrary nested list behaviour will have to be done via a jsonpath selector. It's too difficult to express 'I want an element to be present, somewhere' with the current syntax or without adding a load of sentinels (which like you say would be bad)

francisr8:
Please approve this!

michaelboulton:
merged into 1.0 branch

Issue #163, opened 2018-07-27
cannot install from other pip_index


gocarlos:
When trying to install this package with

```bash
pip3 install -i https://devops.myURL/artifactory/api/pypi/pypi-all/simple -r requirements.txt
```
with requirements.txt
```
pytest
tavern[pytest]
```
one gets the following error: 
```
Collecting pytest (from -r requirements.txt (line 1))
  Downloading https://devops.myURL/artifactory/api/pypi/pypi-all/packages/77/64/3a76f6fbb0f392d60c5960f2b2fbad8c2b802dada87ca6d1b99c0083a929/pytest-3.6.3-py2.py3-none-any.whl (195kB)
Collecting tavern[pytest] (from -r requirements.txt (line 2))
  Downloading https://devops.myURL/artifactory/api/pypi/pypi-all/packages/83/9f/419752827e66422a144597a40f4d53b1805f6af57ee64006f547bf1f89a0/tavern-0.14.4.tar.gz (42kB)
    Complete output from command python setup.py egg_info:
    Download error on https://pypi.org/simple/pytest-runner/: [Errno 99] Cannot assign requested address -- Some packages may not be found!
    Couldn't find index page for 'pytest-runner' (maybe misspelled?)
    Download error on https://pypi.org/simple/: [Errno 99] Cannot assign requested address -- Some packages may not be found!
    No local packages or working download links found for pytest-runner
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-_4ygw1zi/tavern/setup.py", line 87, in <module>
        "tests": TESTS_REQUIRE
      File "/home/one/.local/lib/python3.5/site-packages/setuptools/__init__.py", line 130, in setup
        _install_setup_requires(attrs)
      File "/home/one/.local/lib/python3.5/site-packages/setuptools/__init__.py", line 125, in _install_setup_requires
        dist.fetch_build_eggs(dist.setup_requires)
      File "/home/one/.local/lib/python3.5/site-packages/setuptools/dist.py", line 514, in fetch_build_eggs
        replace_conflicting=True,
      File "/home/one/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 773, in resolve
        replace_conflicting=replace_conflicting
      File "/home/one/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 1056, in best_match
        return self.obtain(req, installer)
[...]

PR #165, opened 2018-07-30
Use `install_requires` for the pytest-runner

benhowes:
Possible fix for #163 

edit: @michaelboulton I'm really confused as to how this was working at all given that it's now complaining about all these other dependencies being missing?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,9 +55,11 @@ class DeployPypi(Command):
             [args],
             msg="Uploading package to pypi")
 
-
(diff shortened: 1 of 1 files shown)


michaelboulton:
install_requires is in setup.cfg, not setup.py. test_requires and setup_requires are only in setup.py because of limitations of the way setuptools/distutils works.

michaelboulton:
Fixed this in master, will deploy shortly

PR #167, opened 2018-08-04
Initial timeout implementation

### LINKED ISSUE #164: Support for a timeout parameter for requests
I would like to be about to specify a timeout parameter for each of my requests, this would be useful for endpoints that must return a given timeout. https://github.com/taverntesting/tavern/blob/2ff7f97e3c705b9a935ab05aef52545a85b332cb/tavern/_plugins/rest/request.py#L64-L67 seems like a good place to slot it in with a possible default of `None`. I wouldn't mind making a PR for this if this route seems like a good idea.

michaelboulton:
closes #164


benhowes:
Looks good! Clearly just a minor testing problem to sort out.

Needs some additions to the docs, which should link to the correct part of the requests documentation IMHO.

We need to get a CHANGELOG.md setup in this repo too, perhaps this could be the first thing?

michaelboulton:
This is released in 0.18.0

PR #177, opened 2018-08-20
Feature/new hooks

michaelboulton:
Adding some basic hooks which could be expanded further down the line. After thinking about it, I think this is much better than having a fixture.

At the time of writing this is just a simple thing which prints a response (not fully implemented), some other obvious ones are:

- called with request parameters
- called before each stage
- called after each stage
- ?

They will all basically just be called with the test name, stage name, stage number, and then a dictionary of any extra relevant information (the schema of which is undefined).

This is not the same as https://github.com/taverntesting/tavern/pull/115, these hooks are always called


benhowes:
I'm trying to understand the use cases and how someone would use this? Can you add multiple before after hooks?

From the code I can see you mention being able to do timing, and there's some stuff about logging. Are there other things which could be done too? Some docs would be helpful. I'd be happy to write them if you can expand on it a bit here.

michaelboulton:
The use case is mainly to provide people with a way to implement their own ways of dealing with the test results. For example pytest-html lets you create a html report from pytest tests, but won't allow you to fully express the concept of the headers/response body/status code and dump it to a machine readable format.

If we add something like a `pytest_tavern_stage_failed` hook then people could make it dump out an xml document describing exactly what went wrong (eg, incorrect header, error in body, etc.), or if it's being used for uptime monitoring it could post to a server which displays the status of the server (500 response, timeout, etc).

On top of that, there is some stuff in the tests like `delay_after` and `delay_before` which is baked into each test, this is an attempt to make it so that this kind of stuff doesn't need to be handled in the core of the library and can be provided by third party plugins.

kvenkat88:
1.It would be good if we provide Unicode string object validator like !anyunicode or something related to that...As far I know from documentation there is no option available.Is there any feature like this implemented.I am using external function for validating this.
2.It would be good if usefixtures feature with extra_kwargs to pass the arguments to implement our requirements in the runtime.As of now fixtures doesn't have this feature correct me if I am wrong.I am not mentioning fixtures with parametrization (know pytest have issues with fixture parametrization)
3.I read the issue comments somewhere and observed $ext(external function) is going to be replaced with fixture.It would be good that replacement fixture/hook have direct object return support instead of saving and use it in later.

michaelboulton:
https://github.com/taverntesting/tavern/pull/415

PR #182, opened 2018-08-29
Add retries

jgroom33:


KPilnacek:
Hi @jgroom33 and everybody else 🙂,

I wanted to do the PR myself today, so thanks for support with this.

Anyway, in our company we are tried `tavern` and we are very happy with it. However, we couldn't figure out a way how to do polling. Thus we put together this PR, which allows to retry a test a specified number of times. If no `retry` is specified or is set to 1, the original implementation is used.

I hope it fits well within this framework and helps other people to use it.
The last thing that bugs me is that I couldn't find a place to document this feature. FYI: I'm willing to write it.

Feel free to comment and review.
With regards
Krystof

Birne94:
I think having a retry value of `1` indicating no retries being performed is a little bit confusing. If I understand correctly, this parameter controls the total amount of tries attempted before failing, not just the amount of retries.

I suggest having a default value of `0` which indicates no additional retries being performed.

KPilnacek:
Thanks for the remark. Please see [the fixup](https://github.com/taverntesting/tavern/pull/182/commits/407b255c56ef6571353838a162d03450efb06424).

In the end, inspired by [requests](http://docs.python-requests.org/en/master/api/?highlight=max_retries), I even changed the name to `max_retries`, in order to be more descriptive.

michaelboulton:
I don't think that running the delay after each retry is what should happen - the delay should only happen after the stage has finished. This should be removed or another variable (eg `retry_interval`) should be added.

KPilnacek:
@michaelboulton I was thinking the same, however, `delay_before` is run before each retry so I wanted to be consistent and make as few changes in the `core.py` as possible.

Please see possible solution in https://github.com/KPilnacek/tavern/commit/f06608e7eea66597eb29e1429f530407bf0f5808 or IMO a bit more intrusive but cleaner https://github.com/KPilnacek/tavern/commit/84f70854c192f35fa157cb153eea9d27cffb17ec

jgroom33:
I think using the delay_after property is logical.  Unless, there is a use case for different values in delay_after and retry_interval.
Generally, I see the existing use of delay_after as meant to be:
"Some period of wait time is necessary, but the exact amount is not know, so fail if the wait time exceeds delay_after"  This results in wait times that are higher than might be necessary. (i.e. the system is actually ready prior to the time defined in delay_after but a large time was specified because it *could* take longer in some circumstances)

michaelboulton:
Sorry about the late reply - I can see why doing the delay_after is useful if you are waiting for a service to be ready, for example. It hsouldn't interfere with existing behaviour anyway.

KPilnacek:
No, problem :-). I'm glad to see it merged.

Thanks

Issue #211, opened 2018-11-21
Parametrize without combinations.


rvanderwall:
I am trying to create a parameterized test but don't want the combinations. for example:

```
-parametrize
    key: p1
    vals:
        - a
        - b
-parametrize
    key: p2
        -x
        -y
```
This gives 4 tests, (a,x), (a,y), (b,x), (b,y)but I only want 2 tests (a,x), (b,y)

Another use case is that the pass criteria changes with each parameter.  For example:
```
key: input
vals:
    - 1
    - 3
key: double
vals:
    - 2
    - 6
```
Is that possible? thanks.

benhowes:
@michaelboulton I've had a dig through the docs and tests and can't see any examples of us using parametrize with a list of dictionaries as the `vals`. Should that work? I can add a test/docs if so

rvanderwall:
Each test input may have a different output.  For example, passing in a 2 returns a 4 while passing in a 3 returns a 6.  I want to parameterize tests something like this: (in,out) = [(2, 4), (3,6)] and the test would use {in} on the request and {out} on the response.  Hope that makes sense.

michaelboulton:
I'm assuming you mean like this

```yaml
test_name: parametrise input/output

marks:
  - parametrize:
      key: input
      vals:
        - 1
        - 2
  - parametrize:
      key: output
      vals:
        - 2
        - 4

stages:
  - name: double a number
    request:
      url: "http://example.com/double"
      method: POST
      json:
        value: "{input}"
    response:
      status_code: 200
      body:
        value: "{output}"
```
?

In which case, I think that should work as it is - have you run into any issues with it?

rvanderwall:
I tried that, but I end up with 4 tests, (1,2), (1,4), (2,2), and (2,4).
This may be a different issue, but I cannot parameterize the method.  Looks like it is hard-coded to valid verbs only.  (if it is a different issue, I'll write up a separate request)

Birne94:
While maybe not as clean of a solution, using helper functions you can achieve the desired results:

```yaml
test_name: parametrise input/output

marks:
  - parametrize:
      key: params
      vals:
        - "1|2"
        - "2|4"

stages:
  - name: double a number
    request:
      url: "http://example.com/double"
      method: POST
      json:
        $ext:
          function: utils:create_request
          extra_kwargs:
            param: "{params}"
    response:
      status_code: 200
      body:
        $ext:
          function: utils:verify_response
          extra_kwargs:
[...]

Issue #191, opened 2018-09-22
Make it so you can 'save' more complicate blocks from a response?


michaelboulton:
- https://github.com/taverntesting/tavern/issues/189
- https://github.com/taverntesting/tavern/issues/84#issuecomment-398709340
- https://github.com/taverntesting/tavern/issues/184

This would just apply to the body. Obviously apart form backwards compatability issues there would need to be checks on not using json in headers, etc. This might just not be possible, but I'll try to take a look.

PR #212, opened 2018-11-22
Feature/parametrize combinations

michaelboulton:
#211 

Need to fix Python 2 later


michaelboulton:
Can add documentation when this is merged - there are examples in the code, but basically if the 'key' is a list, then the values must be a list of the same length, which are parametrised in order according to the key names in the list.

tavern/schemas/tests.schema.yaml
```diff
@@ -248,6 +249,11 @@ mapping:
                   - type: str
                   - type: int
                   - type: float
+                  - type: seq
```
Birne94: What is the rationale behind limiting the type for the parameter values? If tavern allowed here, we would be able to do parametrisation of multiple parameters by simply passing objects:

```yaml
test_name: Make sure server doubles number properly

marks:
  - parametrize:
      key: params
      vals:
        - input: 5
          output: 10
        - input: 6
          output: 12

stages:
  - name: Make sure number is returned correctly
    request:
      url: http://localhost:5000/double
      json:
        number: !int "{params.input}"
      method: POST
      headers:
    response:
      status_code: 200
      body:
        double: !int "{params.output}"
```


```diff
diff --git a/tavern/schemas/tests.schema.yaml b/tavern/schemas/tests.schema.yaml
index 02e4ea1..9e7ffa5 100644
--- a/tavern/schemas/tests.schema.yaml
+++ b/tavern/schemas/tests.schema.yaml
@@ -245,9 +245,7 @@ mapping:
                 type: seq
                 required: true
                 sequence:
-                  - type: str
-                  - type: int
-                  - type: float
+                  - type: any
 
   _xfail:
     type: str
```

michaelboulton: Mainly because I would rather keep the amount of validation we have to do to a minimum. If we allowed that, we would still need to still support the old way of passing parametrized variables as well as 
- manually validating the structure of any items in `values` so that people can't pass some arbitrarily complicated structure and wonder why it isn't working
- validating that all items in the `values` have the same keys 
- validating that all items have the same type
Using `any` in the schema is something that I only ever want to do if there is absolutely no other way around the problem, because it introduces a huge amount of extra validation/testing.

If #191 is ever implemented then we can revisit this question, but for now I would rather keep it as simple as possible.

PR #255, opened 2019-02-03
Feature/prospector

michaelboulton:
Replace pylint with prospector


benhowes:
@michaelboulton you had mentioned that you were having a lot of trouble with prospector - did you get around those issues? Are others likely to have those issues when developing on Tavern, or was it just a case of the prospector config?

michaelboulton:
It seems to have been fixed in the latest release of prospector where they changed a lot of the options around.

PR #258, opened 2019-02-05
Fix/parametrized names

### LINKED ISSUE #216: Split pytest file into modules
It's really big and can be split into

- actual hooks and cmdline args
- custom error formatter
- test generation types

Moving it might mess up entry points for people who already have it installed, not sure

michaelboulton:
closes #216


michaelboulton:
closes https://github.com/taverntesting/tavern/issues/269

PR #272, opened 2019-02-15
Adding support to specify client certs

dkmalav:
The changes add support for client cert based authentication and ssl cert verification. Verified all tox tests run without issues. The documentation needs to be updated to reflect this support.


tavern/schemas/tests.schema.yaml
```diff
@@ -156,8 +156,18 @@ schema;stage:
...
+          required: false
+          sequence:
+          - type: str
+            required: true
+          - type: str
+            required: true
+
         verify:
-          type: bool
+          type: any
```
michaelboulton: If you change this to 'any' then you should put some validation in `extensions.py` to make sure that it is a boolean or a string, which are the only types accepted by requests for this keyword

tavern/schemas/tests.schema.yaml
```diff
@@ -156,8 +156,18 @@ schema;stage:
           required: false
           func: validate_timeout_tuple_or_float
 
+        cert:
+          func: validate_extensions
+          type: seq
```
michaelboulton: This will never load extension functions so using `validate_extensions` is wrong here - this can be either a tuple or a string, a extra extension function in `extensions.py` will be required to validate it (there are lots of examples of similar functions in that file)

tavern/_plugins/rest/request.py
```diff
@@ -111,6 +112,9 @@ def add_request_args(keys, optional):
     if "auth" in fspec:
         request_args["auth"] = tuple(fspec["auth"])
 
+    if "cert" in fspec:
+        request_args["cert"] = tuple(fspec["cert"])
```
michaelboulton: See comment below - this can be either a string or a tuple when given to requests so it should handle both eventualities here as well.

michaelboulton:
Thanks for this - I'll document this now.

PR #274, opened 2019-02-16
Add support for specifying cookies in request

### LINKED ISSUE #257: Selecting session cookie not working OK when multiple cookies are returned
Hi,

I have a login webservice returning two cookies as a response in the response header called "set-cookie". One is called X and the other SCISID9944 let's say. 

I have the following yaml set up and expected the authentication to work on the following webservice calls, but it doesn't as tavern seems to pass only a single cookie (and the wrong one, X) to the webservice call. 

```yaml
test_name: Login

stages:
  - name: login
    request:
      url: http://localhost:9003/login
      auth:
        - ****
        - ****
      json:
        username: ppuser
      method: POST
    response:
      status_code: 200
      cookies:
        - SCISID9944

  - name: get structure metadata/StructureMetadataByCodeGet
    request:
      method: GET
      url: http://localhost:9003/api/v2/structure/1uyd/metadata
    response:
      status_code: 200
      strict: false
```

The login works OK, the second step I get a 401 unauthorized answer. If however I set the cookie manually in the yaml Cookie: cookiestring -> then the second call works. 
I'm not sure what I'm doing wrong here. 

THanks in advance for your reply.

michaelboulton:
closes #257


benhowes:
Looks like this could do with some documentation?

It seems that the behaviour is that it sends all the cookies in the session (which requests stores) unless you tell it to send particular cookies, in which case it only sends those ones for that particular request?

Do we want a way to do all except a particular cookie? Or a way to send no cookies? e.g. `cookies: False`

michaelboulton:
I keep forgetting the docs are now in this repo too. sending no cookies would be done via `cookies: []`. By default it would just do whatever requests does, which judging from #257 seems to be a bit weird

michaelboulton:
Documentation added @benhowes

Issue #188, opened 2018-09-21
change recurse_access_key to use jmespath, fallback to old behaviour, and add deprecation warning


michaelboulton:
#119 

This doesn't need to be removed in 1.0, but it can be deprecated

michaelboulton:
Done with #289

PR #289, opened 2019-02-27
Add JMESpath selector

### LINKED ISSUE #273: Add Json path / jmes path selector
This would be an extra key in the response and mqtt_response block:

```yaml
    response:
      status_code: 200
      body:
        a: 1
        b: 2
        c: !anything  # a very complicated object
      jsonpath:
        - path: "c.f.h[1]"
          save_as: c_f_h_1  # optional
          value: a_nested_thing
```
This would:
- simplify saving parts of the request
- simplify saving from complicated objects
- allow parametrizing of expected response data (eg `path: "{deep_object_path}"`)
- triage various issues to do with selecting from nested lists (https://github.com/taverntesting/tavern/issues/238, https://github.com/taverntesting/tavern/issues/133, https://github.com/taverntesting/tavern/issues/262, https://github.com/taverntesting/tavern/issues/265, https://github.com/taverntesting/tavern/issues/267, https://github.com/taverntesting/tavern/issues/133)

@benhowes  ?

michaelboulton:
closes #273 

As mentioned in the documentation, this adds the ability to use a more complex query to search for things in , for example, nested lists, which is an issue that comes up constantly.


michaelboulton:
closes https://github.com/taverntesting/tavern/issues/188

michaelboulton:
@benhowes 

This is basically ready to merge, but it is only implemented for HTTP requests. It would only be a couple of hours of extra work to add it for MQTT, but it feels like a lot less people use it so we can just add it later if somebody asks for it?

benhowes:
I'm worried that we're loosing the cleanliness that makes Tavern so easy to read and write, because this adds a second way to match values and a second way to save values. That of course can lead to confusion and resulting support requests.

Do we perhaps want to make every path a jmes path in a breaking change for 1.0?

michaelboulton:
I thought this would make every search path a jmes path, I did add deprecation warnings so that any use of the 'old' style paths should raise a warning too.

I agree that this is adding another way of validating things which makes it a bit messier and less 'opinionated', but there is a lot of issues about matching things in lists which is impossible to express at the moment (tried to fix it but ran into issues in that other PR https://github.com/taverntesting/tavern/pull/162#issuecomment-414304895). We could add something into the documentation to say that it should only be used when the 'old' style of validating responses absolutely cannot do what you want, because it is a lot easier to read a block of text rather than some weird DSL

michaelboulton:
Going to close this, now we have the jmespath selector for saving and locating things then this is superfluous

willsthompson:
@michaelboulton Can the current jmespath selector implementation be used to validate a value (i.e. how `expected` appears to work in this PR)? It seems like JMESPath values can only be saved and reused in later requests, and not validated directly. I gave up and instead used `tavern.testutils.helpers:validate_content`, but it would be nice if it could be done without the external helper.

PR #295, opened 2019-03-02
Fix/mqtt payload match

### LINKED ISSUE #293: Type tokens not working in mqtt block
I used the example from the documentation (Sidenote: it uses mqtt as a label instead of paho-mqtt so it should be updated). I made the mqtt-publish and mqtt response both look for ping and it worked. If I changed the mqtt-response to !anything I get the error below. I tried the same thing with json and had the same errors.

### Code:
```yaml
# test_mqtt.tavern.yaml
---

test_name: Test mqtt message response
paho-mqtt:
  client:
    transport: tcp
    client_id: tavern-tester
  connect:
    host: 10.1.0.67
    port: 1883
    timeout: 3

stages:
  - name: step 1 - ping/pong
    mqtt_publish:
      topic: test
      payload: ping
    mqtt_response:
      topic: test
      payload: !anything
      timeout: 5
```

### Error Output (edit: removed < from before "Tavern" on top and bottom line. Was causing the line not to show up).
Got unexpected payload on topic 'test': 'ping' (expected 'Tavern YAML sentinel for anything>')
Message not received after 4 seconds
Expected 'Tavern YAML sentinel for anything>' on topic 'test' but no such message received

michaelboulton:
closes #293


tox-integration.ini
```diff
@@ -42,8 +42,8 @@ commands =
 
     mqtt: tavern-ci --stdout test_mqtt.tavern.yaml
     mqtt: python -c "from tavern.core import run; exit(run('test_mqtt.tavern.yaml', pytest_args=[]))"
-    mqtt_failures: tavern-ci --stdout test_mqtt_failures.tavern.yaml
-    mqtt_failures: python -c "from tavern.core import run; exit(run('test_mqtt_failures.tavern.yaml', pytest_args=[]))"
+    mqtt: tavern-ci --stdout test_mqtt_failures.tavern.yaml
```
benhowes: Will this work with duplicate keys?

michaelboulton: It's ot a 'key', its just saying which tests to run under which environment

Issue #307, opened 2019-03-15
pyyaml deprecation warning


michaelboulton:
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
    schema = yaml.load(schema_file)

michaelboulton:
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
    self._loaded[schema_filename] = yaml.load(sfile.read())

PR #309, opened 2019-03-16
Remove deprecation warnings for pyyaml and collections

michaelboulton:
closes https://github.com/taverntesting/tavern/issues/307


michaelboulton:
Also adds a warning to say that Python 2 support will be dropped

tavern/testutils/pytesthook/item.py
```diff
@@ -29,13 +31,23 @@ class YamlItem(pytest.Item):
...
         super(YamlItem, self).__init__(name, parent)
         self.path = path
         self.spec = spec
 
         self.global_cfg = {}
 
+        if sys.version_info < (3, 0, 0):
+            if not YamlItem.py2_warned:
+                warnings.warn(
+                    "Tavern will drop support for Python 2 in a future release, please switch to using Python 3",
```
benhowes: Our trigger for dropping will be when pytest stops supporting python 2 right? should we mention that and/or link to the pytest roadmap https://docs.pytest.org/en/latest/py27-py34-deprecation.html

michaelboulton: Done

Issue #308, opened 2019-03-16
Variable error logging doesn't handle escaped curly bracket, !raw, and comments


MichielVanderlee:
`tavern/testutils/pytesthook/error.py` `ReprdError.toterminal().read_relevant_lines()`
Does not ignore lines starting with `#` or containing `!raw`. This causes errors when using curly brackets in comments.

Version 0.25.0

MichielVanderlee:
Took a while to figure out exactly what's going on here.
If a test fails, i.e.: response body doesn't match, then it reads the `yaml` file line by line to extract all variables via this regex: `r"(?P<format_var>{.*?})"`.

There are a few issues with this approach:
1. We read and parse comments i.e.: `# This will match the {regex}`
2. We read and parse `!raw` values. `raw_value: !raw "{uhoh}"`
3. We match incorrectly on escaped curly brackets, i.e.: ` escaped_value: "{{I'm not a variable}}"`

michaelboulton:
Should be fixed by the above commit

PR #311, opened 2019-03-16
Issue #308

MichielVanderlee:
When we read the line, strip any comments from the end.
Regex: 
* Only look for variables in lines that don't have `!raw`
* Only match variables encapsulated by a single curly bracket pair.
i.e: match `{var}` and not `{{var}}`


MichielVanderlee:
Fixes issue #308

michaelboulton:
Thanks for fixing this, I see you noticed the comment in there saying it needed fixing as well!

Issue #245, opened 2019-02-02
Change loading configs to catch multiple documents error


michaelboulton:
Loading global/local config files should only expect one document per file and should raise an error if there is more than one document

eg 

```yaml
# cfg.yaml
---
variables:
  a: 1
---
variables:
  b: 1
```

PR #317, opened 2019-03-19
Fix core tests to not recursively run tavern tests and add more tests

michaelboulton:
closes https://github.com/taverntesting/tavern/issues/245


michaelboulton:
This is affecting development on every other branch (and master due to the broken dependency) so I'm force merging it

PR #322, opened 2019-03-23
Feature/add missing docs

michaelboulton:


docs/source/basics.md
```diff
@@ -1073,6 +1074,162 @@ tavern-global-cfg=
...
+```yaml
+---
+
+test_name: Test authenticated /hello
+
+includes:
+  - !include auth_stage.yaml
+
+stages:
+  - name: Unauthenticated /hello
```
benhowes: Maybe get rid of this first request. I know it's completely correct, but it makes spotting the usage of the ref less obvious

PR #324, opened 2019-03-24
Change 'save' selector to use JMES path

michaelboulton:
Part of #289 

In conjunction with #119


docs/source/basics.md
```diff
@@ -121,9 +121,17 @@ This can be saved into the value `first_val` with this response block:
 response:
   save:
     body:
-      fourth_val: thing.nested.0
+      first_val: thing.nested[0]
 ```
 
+The query should be defined as a JMES query (see http://jmespath.org/
+for more information). In the above example, this essentially performs
+the operation `json["thing"]["nested"][0]`. This can be used to perform
```
benhowes: I started drafting some changes to "In the above example, this essentially performs the operation `json["thing"]["nested"][0]`" but I think in reality it's probably best to leave it out, because it breaks the flow of the paragraph. Perhaps add a yaml comment with `# The python equivalent of this selector would be json["thing"]["nested"][0]`

tavern/_plugins/rest/response.py
```diff
@@ -1,7 +1,6 @@
 import json
-import traceback
 import logging
-import copy
+import traceback
```
benhowes: Not used?
```suggestion
```

tavern/util/dict_util.py
```diff
@@ -56,26 +52,89 @@ def format_keys(val, variables):
...
+        except (IndexError, KeyError) as e:
+            raise_from(
+                exceptions.KeySearchNotFoundError(
+                    "Error searching for key in given data"
+                ),
+                e,
+            )
+
+        # If we found a key using 'old' style searching, which will be deprecated
+        warnings.warn(msg, FutureWarning)
```
benhowes: Perhaps we can call warning.warn inside the `_recurse_access_key` function because that will be cleaner when we remove it?

tavern/util/jmespath_util.py
```diff
@@ -0,0 +1,35 @@
...
+
+    msg = "JMES path '{}' not found in response".format(query)
+
+    if actual is None:
+        raise exceptions.JMESError(msg)
+
+    if expected is not None:
+        # Reuse dict util helper as it should behave the same
+        check_keys_match_recursive(expected, actual, [], True)
+    elif not actual and (not actual == expected):
```
benhowes:
```suggestion
    elif not actual and not (actual == expected):
```
might be better if you're looking to clarify the order of operations?

tavern/util/dict_util.py
```diff
@@ -56,26 +52,89 @@ def format_keys(val, variables):
...
+
+    try:
+        from_jmespath = jmespath.search(query, data)
+    except jmespath.exceptions.ParseError:
+        logger.warning("Error parsing JMES query - %s", msg, exc_info=True)
+        from_jmespath = None
+
+    # The value might actually be None, in which case we will search twice for no reason,
+    # but this shouldn't cause any issues
+    if from_jmespath is None:
```
benhowes: Ah crap. This is going to warn if the value is None.

I'm sure you'd already seen that we cannot distinguish between None and not present with search (https://github.com/jmespath/jmespath.py/issues/113).

Perhaps we should only do the old search if the path is not a valid JMESpath? or will that cause `foo.0.bar` to silently fail?

tavern/_plugins/rest/response.py
michaelboulton: It's used in a couple places in that file

tavern/util/dict_util.py
michaelboulton: Would that not spam the warning a lot of times?

This function will probably be completely removed when we do switch to just using jmespath anyway

tavern/util/dict_util.py
michaelboulton: You're right in that it would break the current behaviour. It is a bit of a hole in the spec but I can add documentation saying to use that 'contains' function if you need to check for null?

docs/source/basics.md
michaelboulton: I think that sentence only makes sense of the context of it being a JMES path query though

PR #325, opened 2019-03-25
Make advanced example work outside UTC

rogerrohrbach:
The example server for the multi-stage test example was signing the
token in the local timezone. jwt.decode expects UTC, so if the local
time was offset from that, jwt.ExpiredSignatureError was raised.


rogerrohrbach:
The failed test exercises code I did not touch.

michaelboulton:
Thanks for noticing this!

PR #343, opened 2019-04-20
Allow any value to be passed as json body

### LINKED ISSUE #342: Support for RFC 7159 - specifically single value JSON responses
In issue #120 handling of single string JSON responses was requested, but denied on grounds that RFC 4627 doesn't permit this. However RFC 7159 obsoletes RFC 4627, and the JSON grammar allows all valid JSON values.
Specifically: 
false / null / true / object / array / number / string

I would like to raise this issue again, on grounds that a more current RFC for JSON does define a string (with quotes around it) as being valid.

My hack for handling strings:
\-\-\- extensions.py       2019-04-17 16:17:50.072416800 -0300
\+\+\+ patch/extensions.py 2019-04-03 15:08:36.872055300 -0300
@@ -326,6 +326,10 @@
     """
     validate_extensions(value, rule_obj, path)

\+    # RFC 7159 says we can have a string, int or boolean, and dict and list
\+    if isinstance(value, str):
\+        return True
\+
     if not isinstance(value, (list, dict)):
         raise BadSchemaError("Error at {} - expected a list or dict".format(path))

michaelboulton:
closes #342


benhowes:
@michaelboulton I've pushed some tests for what I think this is trying to do. I'm not completely sure if you're trying to support any type in the request and response?

michaelboulton:
Yes, in general any single value should be allowed as part of a request or response

michaelboulton:
One thing I just found out is that this does break the 'ext' functions in responses because you can't put a single value _and_ a key:value mapping in it.

I'm fine with just saying this is a limitation of using a single value response because it's such a simple check anyway, and it's not breaking existing behaviour. Thoughts @benhowes ?

calebreid:
Thank you for all the work you are doing to develop Tavern, we are enjoying using it. And this feature would make it even more helpful for our testing, do you have an idea of when it might be merged?

PR #360, opened 2019-05-23
Update basics.md: fix "welcome_message" variable

pmull:
I think that "tavern.request_vars.json.message" must be "tavern.request_vars.json.welcome_message".


michaelboulton:
Thanks for spotting, It would be useful for me to go through the documentation and clear out things like this at some point (I did just notice another typo in there myself!)

PR #379, opened 2019-06-26
use paho-mqtt instead of mqtt 'keyword'

TanjaBayer:
It seams like running tests with 'mqtt' keywords is not (no more?) running, instead paho-mqtt needs to be used
Updating the docu according to that will help new users


michaelboulton:
Thanks, it was `mqtt` a long time ago but the docs never got changed.

Issue #382, opened 2019-06-29
Allowing `files` and `data` in a single request


swiatekm:
Is there a reason this is not allowed? As far as I can tell, `requests` handles this case exactly as one would expect:
```python
import requests, io
url, data, files = "https://example.com", {"key": "value"}, {"file": io.BytesIO(b"content")}
prepared_request = requests.Request('POST', url, data=data, files=files).prepare()
print('{}\n{}\n{}\n\n{}'.format( 
  '-----------START-----------', 
  request.method + ' ' + request.url, 
  '\n'.join('{}: {}'.format(k, v) for k, v in request.headers.items()), 
  request.body, 
))
```
Giving:
```
-----------START-----------
POST https://example.com/
Content-Length: 236
Content-Type: multipart/form-data; boundary=0800bc8d540aef3694c7294488bc66ba

b'--0800bc8d540aef3694c7294488bc66ba\r\nContent-Disposition: form-data; name="key"\r\n\r\nvalue\r\n--0800bc8d540aef3694c7294488bc66ba\r\nContent-Disposition: form-data; name="file"; filename="file"\r\n\r\ncontent\r\n--0800bc8d540aef3694c7294488bc66ba--\r\n'
```
Enabling this in tavern appears to come down to relaxing this condition: https://github.com/taverntesting/tavern/blob/7cf8c4033669b81122e00768c95eaa7fdde45494/tavern/_plugins/rest/request.py#L77
and updating the relevant tests.

I have some integration tests that do this running on a fork with this change, and am willing to submit a PR, assuming there isn't some other reason this isn't allowed.

PR #384, opened 2019-07-02
Allow sending data and files in a single request

swiatekm:
Requests handles this case correctly, so I don't see any reason for tavern to forbid it. And I find it to be a very normal use case for the library, as APIs often include endpoints of this form. In particular, image uploads alongside form text data are fairly typical in my experience.

This fixes https://github.com/taverntesting/tavern/issues/382.


swiatekm:
Please note that the CI failure is apparently some kind of recent incompatibility between prospector and pylint. This PR does in fact pass all the unit and integration tests.

vsgiri:
This is not a review. I can however confirm that changes by @swiatekm help with requests that have data alongside files. I can confirm it works for me.

PR #395, opened 2019-08-04
Fix minor inconsistency in doc filenames

sujaymansingh:
The yaml filename to be included was called `stage_auth.yaml` but the
yaml file with tests tries to include a file called `auth_stage.yaml`


michaelboulton:
Good spot, thanks for fixing it

PR #406, opened 2019-08-13
Allow substitution of env vars in included filenames

Chris2048:
Allow substitution of env vars in included filename, such that a run:

```shell
TAVERN_ENVIRONMENT="local" tavern-ci ...
```
Can be used with an include of the for:

```yaml
includes:
  - !include "includes/{TAVERN_ENVIRONMENT}.yaml"
```

### DIFF
--- a/tavern/util/loader.py
+++ b/tavern/util/loader.py
@@ -137,7 +137,7 @@ def construct_include(loader, node):
                 filename
             )
         )
-
+    filename = filename.format(**dict(os.environ))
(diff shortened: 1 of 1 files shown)


michaelboulton:
Is there a reason you want to do it like this rather than just using a different global configuration file for each environment https://tavern.readthedocs.io/en/latest/basics.html#including-global-configuration-files ?

Chris2048:
No, I missed that bit

Issue #230, opened 2019-01-07
[Feature Request] Add "Save" Functionality to MQTT Response


KenStorey:
It would be useful to have the facility to capture received data.

michaelboulton:
This should be covered by #177

PR #409, opened 2019-08-14
ext and save function improvements

michaelboulton:
Started as trying to just clean up the ext functions, but then I realised how the existing external function implementation was a bit over engineered and had a lot of holes in the implementation

This breaks quite a few bits of the language so it's a 1.0 change

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -92,8 +92,6 @@ tags
 .pytest_cache
 
 .bumpversion.cfg
-.cache
-.coverage
 dist
 docker
 docs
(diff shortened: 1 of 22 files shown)


michaelboulton:
Going to change to not be a breaking change but to warn instead, will be broke in 1.0

michaelboulton:
- closes https://github.com/taverntesting/tavern/issues/414
- closes https://github.com/taverntesting/tavern/issues/230
- closes https://github.com/taverntesting/tavern/issues/191

Issue #411, opened 2019-08-15
Testing Raw JSON in Response Body


dgrahn:
I have the following test, which works.

```
---
test_name: ...

stages:
  - name: ...

    request:
      url: ...
      method: GET
    
    response:
      status_code: 200
      body: { "hello": "world" }
```

I would like to write the test like this.

```
---
test_name: ...

stages:
  - name: ...

    request:
      url: ...
      method: GET
    
    response:
      status_code: 200
      body: >
        {
            "hello": "world"
        }
```

But it gives me the error.

```
2019-08-15 14:20:26,724 [ERROR]: (tavern.util.dict_util:53) Failed to resolve string [{ "hello": "world" }
```

dgrahn:
I'm thinking that this is related to the variable expansion, but I can't find a way to bypass it.

p-jaya:
Is there a reason why you don't just write this:

```
...
response:
  body:
    hello: world
```

?

dgrahn:
@p-jaya Because I want to be able to copy very long responses in in raw JSON without converting them over to YAML.

I was able to make some progress. The following code gives a new error.

```
      body: >
        {
            "hello": "world"
        }
```

The new error.

```
2019-08-15 14:33:34,025 [ERROR]: (tavern.response.base:47) Value mismatch in body: Type of returned data was different than expected (expected = '{                                                                                         "hello": "world"
}
' (type = <class 'tavern.util.dict_util.format_keys.<locals>.InnerFormattedString'>), actual = '{'hello': 'world'}' (type = <class 'dict'>))
```

Birne94:
Have you tried omitting the `>`?

This should work as expected since json is a subset of yaml:

```yaml
---
test_name: ...

stages:
  - name: ...

    request:
      url: ...
      method: GET
    
    response:
      status_code: 200
      body:
        {
            "hello": "world"
        }

```

michaelboulton:
As noted above, if you want to include some JSON you can just paste it right in the test because JSON is also valid YAML. If you want to use an external file, you can include it directly: https://tavern.readthedocs.io/en/latest/basics.html#including-external-files

dgrahn:
I don't know why it worked this time around, but I had definitely tried it before. Thanks.

PR #412, opened 2019-08-15
Allow raw JSON in response body.

dgrahn:
Adding this check here allows raw json in the response body. Fix for #411


dgrahn:
I think this is failing because I put an import inside an if statement. It was just for conciseness.

michaelboulton:
Assuming this https://github.com/taverntesting/tavern/issues/411#issuecomment-522058032 solves your original issue, is this PR required any more?

Decoding the json value which has already been decoded will also break existing behaviour.

Issue #270, opened 2019-02-13
Is there any way to print the JSON response?


varayal:
I'm just trying to get the whole JSON response after running the test for evaluating after execution.

Any thought about how to do it?

varayal:
by the way, I tried with $ext and I got the following error

```
>   ???
E   tavern.util.exceptions.BadSchemaError: Couldn't load utils:test_function

<string>:1: BadSchemaError
```

➜  tree
.
├── includes.yaml
├── test_createOrder.tavern.yaml
└── tfunctions
    └── utils.py

```
      save:
        $ext:
          function: utils:test_function
```

```
# utils.py
def test_function(response):
    return {response.json()}
```

srinivasgoud-n:
Please let me know which command you are using to run the tests

varayal:
tavern-ci test_test01.tavern.yaml

felipehernandez:
This may be caused by Tavern not finding your file. I got it solved by:
- Adding __init__.py to your tfunctions folder and running `PYTHONPATH=$PYTHONPATH:. py.test ./`
- Running PYTHONPATH=$PYTHONPATH:tfunctions py.test ./
This will add your py files to the Python Path and make them visible to Tavern

michaelboulton:
https://tavern.readthedocs.io/en/latest/basics.html#after-every-response

Issue #378, opened 2019-06-25
Printing the variables passed in yaml


devang-s:
Suppose I am passing trace-id using the !uuid or using referring fixtures to pass values into the request(passing with {fixture-name})and I want to see the values passed.

Is it possible to do so?

michaelboulton:
This isn't possible easily, but you can add a hook that will run before each stage which will give you the test dictionary and the variables (see here: https://tavern.readthedocs.io/en/latest/basics.html#before-every-test-run)

Issue #219, opened 2018-12-19
Possibility of print/output the response of external function return objects support in console


kvenkat88:
Is there any possibility to include the support of including the print statements in the .yaml test files. It would be good for interpretation. For example,

```yaml
stages:
  - name: Launch /posts API endpoint 
    request:
      url: "{server_url:s}"
      method: GET
    response:
      status_code: 200
      body:
        $ext:
          function: eva_ask_test.test_utilities.helpers:json_validator #Have to print the output of this one
```

PR #415, opened 2019-08-17
Add hook which runs before every test

### LINKED ISSUE #113: Plugin config through global config

michaelboulton:
There will be more hooks added to do more stuff in future but this one should provide at least some kind of a base to build the other ones on (see https://github.com/taverntesting/tavern/pull/177)

This just adds a hook which is called before every test (As according to the docstring) with the test block itself and any available variables.

Fixtures are meant for getting variables to use in the test, or to parametrize it, etc. Hooks provide more of an insight into the internals of Tavern to do things that are impossible with fixtures.

closes #113

relates to #177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,6 +102,9 @@ jobs:
     - python: 3.7
       env: TOXENV=py37-advanced TOXCFG=tox-integration.ini
       stage: full-tests
+    - python: 3.7
+      env: TOXENV=py37-hooks TOXCFG=tox-integration.ini
+      stage: full-tests
     - stage: deploy
       script: skip
       deploy:
--- a/docs/source/basics.md
+++ b/docs/source/basics.md
@@ -1424,7 +1424,7 @@ values. Say we want to convert a value from an included file to an integer:
 ```yaml
 request:
   json:
-    an_integer: !!int "{my_integer:d}" # Error
+    # an_integer: !!int "{my_integer:d}" # Error
     an_integer: !int "{my_integer:d}" # Works
 ```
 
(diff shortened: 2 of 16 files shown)


michaelboulton:
Requires tests

michaelboulton:
+docs

michaelboulton:
- closes https://github.com/taverntesting/tavern/issues/270
- closes https://github.com/taverntesting/tavern/issues/113
- relates to https://github.com/taverntesting/tavern/issues/114
- relates to https://github.com/taverntesting/tavern/issues/378
- relates to https://github.com/taverntesting/tavern/issues/219

PR #421, opened 2019-08-24
cookie improvements

### LINKED ISSUE #418: MissingCookieError: Specifying Cookies in multistages test
I was following the multistage test [document](https://tavern.readthedocs.io/en/latest/examples.html#multi-stage-tests)

I wrote test as below
```
test_name: Test
stages:
  - name: login user before sending request
    request:
      url: http://localhost:8080/api/user/login
      method: POST
      json:
        email: "test@gmail.com"
        password: "test"
      headers:
    response:
      status_code: 200
      headers:
      save:
        headers:
          session: Set-Cookie

  - name: Get all users
    request:
      url: http://localhost:8080/api/user/list
      method: GET
      headers:
      cookies: ["{session:s}"]
    response:
      status_code: 200
```
We get the  MissingCookieError exception
tavern.util.exceptions.MissingCookieError: Tried to use cookies '['{session:s}']' in request but only had '{}' available

Am I incorrectly storing the cookie as ``` session: Set-Cookie ```
or is the problem in the usage ```  cookies: ["{session:s}"] ``` ?

michaelboulton:
In the case that the name of a cookie may be included ina  variable, or an external config file, but only one cookie should be sent in a request

closes #418

### DIFF
--- a/docs/source/http.md
+++ b/docs/source/http.md
@@ -178,6 +178,31 @@ stages:
         status: access denied
 ```
 
+#### Overriding cookies
+
+If you want to override the value of a cookie, then instead of passing a string
+to the `cookies` block in the request, use a mapping of `cookie name: cookie
+value`:
+
+```yaml
+  - name: Override cookie value
+    request:
+      url: "{host}/expect_cookie"
+      method: GET
+      cookies:
+        - tavern-cookie-2: abc
+    response:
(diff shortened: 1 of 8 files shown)


michaelboulton:
Also adding support for overriding the value of a cookie

Issue #419, opened 2019-08-20
Kafka Plugin


dgrahn:
I'd like to add support for Kafka.

It looks like I can do this by copying the MQTT plugin and modifying it to work with a python-kafka library. I just want to make sure that this is the appropriate process or if you would recommend something else.

pouledodue:
can you do one for Pulsar?

michaelboulton:
The MQTT one is probably the best one to take inspiration from, and if you do run into any issues I can try and help, but is is very unlikely to ever get merged into mainline Tavern because I don't have any experience with Kafka to maintain it in future.

If you do succeed in making one then I will happily put in the README and in the documentation page so that other people can use it though.

PR #438, opened 2019-09-05
Tavern GRPC support

kiril-me:
GRPC client support.

```
test_name: GRPC message echo
grpc: 
  connect:
    host: localhost
    port: 50051
    tls: false

stages:
  - name: Echo text
    grpc_request:
      service: helloworld.Greeter/SayHello
      body:
        name: "John"
    grpc_response:
      body:
        message: "Hello, John!"
```

GRPC client connection is described using `host`, `port`, `tls`. 

On stage level, `grpc_request` create a request with the `body`. `service` is the full GRPC service name. Besides, there is a possibility to change the `host` inside request:
```
- name: Echo text
    grpc_request:
      host: myhost:50052
      service: helloworld.Greeter/SayHello
      body:
        name: "John"
```
Similar to http `response` GRPC plugin has`grpc_response`. 
To verify error code you can use `status` and `detail`:
```
stages:
  - name: Echo text
    grpc_request:
      service: helloworld.Greeter/SayHello
    grpc_response:
      status: INVALID_ARGUMENT
      details: "Name was not provided"
```

The main problem with GRPC what client should know the protobuf message to make request and read the response.

GRPC plugin supports several features to discover protobuf message structure:
1. Provide protos file (https://developers.google.com/protocol-buffers/docs/proto3).
```
grpc:
  connect:
    ...
  proto:
    source: my-proto-folder
```
The tavern will automatically compile protos and start using them.
2. In case you already have Python generated classes (`protoc` output https://developers.google.com/protocol-buffers/docs/pythontutorial). Setup the proto module location.
```
grpc:
  connect:
    ...
  proto:
    module: proto-folder
```
Using proto `source` file actually generate GRPC python file inside the default module (`proto`) folder. But you can change it update module to custome name.
```
grpc:
  connect:
    ...
  proto:
    source: my-proto-folder
    module: py-generated-proto
```

3. The plugin will try automatically to request protobuf structure from the server using reflections (https://github.com/grpc/grpc/blob/master/doc/server-reflection.md). Just skip `proto` section.

To send metadata with request GRPC has `metadata` map.
```
test_name: GRPC message echo
grpc: 
  connect:
    host: localhost
    port: 50051
    tls: false
  metadata:
    user: John
```


michaelboulton:
Sorry for not leaving a comment on this earlier, but as mentioned in https://github.com/taverntesting/tavern/issues/419#issuecomment-524562314 I don't really want to merge something in that I won't be able to maintain. What we could do is put this as a 'tavern-grpc' plugin into a separate repository under the 'taverntesting' github organisation, and add the schema and a 'stub' implementation in Tavern that just raises an error and points to that repository?

michaelboulton:
Just for the record, I've got this into a roughly working state with 2.0 but I don't want to force everyone who uses to Tavern to suddenly rely on all the random grpc libraries, and once I have a good idea of how to make it not rely on those libraries for an optional feature, it's ready to go

PR #446, opened 2019-09-09
Satisfy pytest warning for failed regex match in helpers

cetanu:
Variable becomes False instead of None when there is no match

This addresses warnings produced by pytest when the match fails, which can become excessive if there's lots of tests or lots of retries

```
 PytestAssertRewriteWarning: asserting the value None, please use "assert is None"
```


michaelboulton:
Thanks for that, I just kind of tuned that warning out but it is kind of annoying!

PR #460, opened 2019-10-09
pytest.ini variables

bruj0:
When run directly from tavern-ci the variable for: ```--tavern-beta-new-traceback``` must be ```tavern-beta-new-traceback = True``` in pytest.ini


michaelboulton:
Thanks for spotting that, it's a bit confusing with the underscores/dashes between different usages of flags to pytest

Issue #468, opened 2019-10-23
Enable skipping stages in a multi-stage test


theonlyx:
Looks like the `skip` mark is used to skip the entire stage. Ideally, it is better to be able to skip some steps in the stages. 

On way I can see to accomplish this is to  support a new field such as:
`ignore` and check that step around 
https://github.com/taverntesting/tavern/blob/master/tavern/core.py#L26

If the field is defined and set to true, don't append the step in the `test_stages`. Thoughts?

PR #469, opened 2019-10-23
support ignore field in a stage

theonlyx:
Enable `ignore` of a step in a multiple stages test. 

If the `ignore` field is defined and it's value is `True`, ignore that step.

### DIFF
--- a/tavern/core.py
+++ b/tavern/core.py
@@ -25,6 +25,8 @@ def _resolve_test_stages(test_spec, available_stages):
     test_stages = []
     for raw_stage in test_spec["stages"]:
         stage = raw_stage
+        if stage.get("ignore", False):
+            continue
         if stage.get("type") == "ref":
             if "id" in stage:
                 ref_id = stage["id"]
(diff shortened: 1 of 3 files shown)


theonlyx:
https://github.com/taverntesting/tavern/issues/468

michaelboulton:
Is this any different to the `skip` options which currently exists (https://tavern.readthedocs.io/en/latest/basics.html#skip)?

theonlyx:
> Is this any different to the `skip` options which currently exists (https://tavern.readthedocs.io/en/latest/basics.html#skip)?

`skip` skips the entire test, not for any particular step. For example, if a test contains steps `A`, and `B`. `C`. `skip` cannot be used to skip one step such as `B`.

michaelboulton:
The 'skip' marker skips a whole test, but `skip: True` can also be put in a stage to skip just that stage

theonlyx:
> The 'skip' marker skips a whole test, but `skip: True` can also be put in a stage to skip just that stage

Obviously I overlooked that feature. Thank you!

PR #481, opened 2019-11-19
Fix jmespath not working with falsy responses

### LINKED ISSUE #470: Test throws JMESError when JMES path returns empty list or 0
I am upgrading tavern from 0.26.5 to 0.30.3, and found some tests failed after upgrade.
The new version will throw error whenever the JMES path returns [] or 0
`tavern.util.exceptions.JMESError: JMES path xxx not found in response`

for example, these content validations will never pass:
```
//function: tavern.testutils.helpers:validate_content

//after deleting an organization
- jmespath: "data[?id=='{organization_data_test_create.id:d}']" 
  operator: "equals"
  expected: []

- jmespath: data.pageable.pageNumber
  operator: "equals"
  expected: 0
```
As these validations are necessary, would you please help to change the JMES path checking logic?

michaelboulton:
closes #470

### DIFF
--- a/tavern/testutils/helpers.py
+++ b/tavern/testutils/helpers.py
@@ -3,6 +3,7 @@ import json
 import logging
 import re
 
+import jmespath
 import jwt
 from box import Box
 from future.utils import raise_from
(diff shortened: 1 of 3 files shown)


michaelboulton:
This was due to a change that was made to try and improve jmespath parsing, but then part of it was reverted and the current implementation doesn't really make sense

Issue #642, opened 2021-02-12
tls.cert_reqs option in mqtt does not work


toniwa:
I found an option `tls.cert_reqs` in [mqtt connection options](https://github.com/taverntesting/tavern/blame/master/docs/source/mqtt.md#L77) and tried mqtts/mqtt over wss with no certificate validation(`CERT_NONE`), however, it does not work.

To come to the point, I guess `check_file_exist("cert_reqs")` should be removed: [here](https://github.com/taverntesting/tavern/blob/c1f9da82590e85afd5a4b232963d9539ce08b954/tavern/_plugins/mqtt/client.py#L69-L70)

This is because:
- According to the [schema](https://github.com/taverntesting/tavern/blob/c1f9da82590e85afd5a4b232963d9539ce08b954/tavern/_plugins/mqtt/schema.yaml#L66-L72), `cert_reqs` could be `CERT_NONE`, `CERT_OPTIONAL` or `CERT_REQUIRED`, not a file.
- `check_file_exists("cert_reqs")` try to open `CERT_NONE/CERT_OPTIONAL/CERT_REQUIRED` file but it always raises IOError since `cert_reqs` is not a file.

PR #643, opened 2021-02-12
Fix/remove file check

toniwa:
closes https://github.com/taverntesting/tavern/issues/642


michaelboulton:
Thanks, looks like it was just copy/pasted incorrectly

PR #645, opened 2021-02-18
bump jwt version

djredhand:
Looking to bump the version for some projects I am working on. Thanks!


michaelboulton:
Thanks for submitting, but it looks like this is a breaking API change for pyjwt and it would break backwards compatibility for anyone using the jwt helper. I will update this as part of the next 'major' release

Issue #653, opened 2021-03-15
mqtt over tls with default CA cert


toniwa:
When I tried MQTT over TLS with Root CA Certificate installed by default (no `ca_certs` option), it did not work.

I believe this comes from here: https://github.com/taverntesting/tavern/blob/69e8cf4db7c067f94d3ad480bd8295c1c6c335a6/tavern/_plugins/mqtt/client.py#L48-L51

`enable` option poped from `tls_args` and then 

https://github.com/taverntesting/tavern/blob/69e8cf4db7c067f94d3ad480bd8295c1c6c335a6/tavern/_plugins/mqtt/client.py#L158-L160

`tls_set` is skipped if `enable` option is used without any other tls options because `tls_args` is empty.

---
`enable` option explicitly removed since paho does not require it.
https://github.com/eclipse/paho.mqtt.python/blob/225ab3757f6818ba85eb80564948d1c787190cba/src/paho/mqtt/client.py#L764

michaelboulton:
Fixed in 1.14.1 (Thanks @toniwa )

PR #654, opened 2021-03-15
fix(mqtt): tls options validation

toniwa:
https://github.com/taverntesting/tavern/issues/653


tavern/_plugins/mqtt/client.py
```diff
@@ -155,7 +155,7 @@ def __init__(self, **kwargs):
 
         self._client.on_message = self._on_message
 
-        if self._tls_args:
```
michaelboulton, line 158: Instead of returning an extra value from `_handle_tls_args`, couldn't this just be `if self._tls args is not None`?

PR #701, opened 2021-06-17
Updated quickstart example - issue #700

### LINKED ISSUE #700: Quickstart example not working
Hi! At @Clariteia we have decided to start using Tavern to test the [Minos Framework](https://github.com/Clariteia/minos_microservice_common). We have just noticed the quickstart example does not work, though.

The problem is that version `1.15.0` seems to have JSON strictness enabled by default.

IMO, the cleanest fix is to set the whole JSON so it works out of the box:

```yaml
---
# Every test file has one or more tests...
test_name: Get some fake data from the JSON placeholder API

# ...and each test has one or more stages (e.g. an HTTP request)
stages:
  - name: Make sure we have the right ID

    # Define the request to be made...
    request:
      url: https://jsonplaceholder.typicode.com/posts/1
      method: GET

    # ...and the expected response code and body
    response:
      status_code: 200
      json:
        id: 1
        userId: 1
        title: "sunt aut facere repellat provident occaecati excepturi optio reprehenderit"
        body: "quia et suscipit\nsuscipit recusandae consequuntur expedita et cum\nreprehenderit molestiae ut ut quas totam\nnostrum rerum est autem sunt rem eveniet architecto"
```

Solutions such as adding the `strict` property to the YAML or the command line `-tavern-strict json:off` option work as well, but add unnecessary noise from my point of view.

If you agree I can make the change myself.

albamig:
This PR resolves #700. The following changes have been accomplished:

- Updated quickstart docs with an executable test.
- Added an integration test to prove it works.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,9 @@ Then, let's create a basic test, ``test_minimal.tavern.yaml``:
           status_code: 200
           json:
             id: 1
+            userId: 1
+            title: "sunt aut facere repellat provident occaecati excepturi optio reprehenderit"
+            body: "quia et suscipit\nsuscipit recusandae consequuntur expedita et cum\nreprehenderit molestiae ut ut quas totam\nnostrum rerum est autem sunt rem eveniet architecto"
 
 This file can have any name, but if you intend to use Pytest with
 Tavern, it will only pick up files called ``test_*.tavern.yaml``.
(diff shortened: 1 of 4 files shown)


michaelboulton:
Thanks for fixing this, the readme probably needs another look at

PR #713, opened 2021-09-14
fix: handle response that is a json list

melnaquib:
trying to validate json reponse that is a list currently fails
code converts it to dict to be handled as expected


michaelboulton:
Can you give an example of this failing? Tavern can handle list responses fine, is there a situation where this isn't working?

PR #714, opened 2021-09-23
fix readme.rst uses markdown

looker53:


michaelboulton:
Thanks for that - should probably convert the whole readme back to markdown now pypi can handle it really

PR #719, opened 2021-10-12
fix sample docker-compose file

spanasik:
`image` should be replaced by `dockerfile`


michaelboulton:
Thanks, probably some bad copy/paste on my part

Issue #725, opened 2021-10-20
Stop logging return values from external functions


methsib:
When external function is called, results of calling external function is getting logged and this is not ideal when return value contains sensitive data. Is there a workaround for this? 

If not, we would really like to have that as an feature. So that we can avoid logging sensitive data coming from external function.

Thank you

michaelboulton:
This should be fixed in 1.16.4

PR #726, opened 2021-10-21
Add a decorator to supress result log in extfunctions

methsib:
This change allows to suppress result log when external functions are returning sensitive information. 

Fixes: #725

We run tavern tests within CI/CD pipelines and external functions are used for various purposes such as generating request body, verifying response and generating request headers.

In some cases, we use external functions to generate `Bearer` tokens. These are sensitive information and we do NOT want them to be logged when running the tests in 3rd party CI/CD providers.

Hope this will be useful to reduce the chance of sensitive information leaking via logs.

Please advise if we need to add documentation or tests. And also where to add them as we couldn't see an obvious spot. 

Thank you

### DIFF
--- a/tavern/util/extfunctions.py
+++ b/tavern/util/extfunctions.py
@@ -89,7 +89,11 @@ def get_wrapped_response_function(ext):
     @functools.wraps(func)
     def inner(response):
         result = func(response, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


methsib:
Closing the this PR as the issue is fixed in https://github.com/taverntesting/tavern/pull/727 by changing the log levels.

PR #730, opened 2021-10-31
Bugfix error condition checking in run() example

willsthompson:
The result of `run()` is a [pytest ExitCode class](https://docs.pytest.org/en/latest/reference/exit-codes.html), an Enum subclass which evidently uses C-style conventions where success is indicated by 0, so a successful run in this example will evaluate to `False` in the conditional and incorrectly print the error. There are a few ways to fix it, but IMO the clearest is to use the Enum class directly.


michaelboulton:
Thanks, those examples are several years old and haven't got a lot of attention

PR #736, opened 2021-11-25
Lock pyjwt to 1.7.1 in component example

hamx0r:
Later versions of PyJWT raise exceptions like `AttributeError: 'str' object has no attribute 'decode'`.   Locking to earlier version allows existing syntax in Tavern examples to work in containerized servers.


michaelboulton:
Thanks, I really need to set up some way of managing dependencies in one place to stop this happening

PR #741, opened 2021-12-01
Allow bools in parameterized values

hamx0r:
The following `marks` entity use to cause error because `vals` couldn't contain Bools (or lists with Bools in them)

```
  - parametrize:
      key:
        - use_this
      vals:
        - True
        - False
```
This didn't work either

```
  - parametrize:
      key:
        - use_this
        - use_that
      vals:
        - [True, True]
        - [True, False]
        - [False, False]
        - [False, True]      
```

This code change allows the above code to work.

### DIFF
--- a/tavern/schemas/tests.schema.yaml
+++ b/tavern/schemas/tests.schema.yaml
@@ -295,11 +295,13 @@ mapping:
                 required: true
                 sequence:
                   - type: str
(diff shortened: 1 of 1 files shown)


michaelboulton:
Thanks, must have missed that in the schema earlier.

PR #747, opened 2021-12-12
Add simple way of running a function before/after stages

### LINKED ISSUE #114: Add support for before/after test stage actions

michaelboulton:
#114

simple way of running things before/after each stage, optionally taking the response as a 'yield' argument

should also be changed to allow specifying it at the test level and the 'global' level in the same way that strictness is

### DIFF
--- a/scripts/smoke.bash
+++ b/scripts/smoke.bash
@@ -5,7 +5,9 @@ set -ex
 PYVER=39
 
 tox -c tox.ini        \
-  -e py${PYVER}flakes \
+  -e py${PYVER}flakes
+
+tox -c tox.ini        \
   -e py${PYVER}       \
   -e py${PYVER}-pytest6       \
   -e py${PYVER}black  \
--- a/tavern/core.py
+++ b/tavern/core.py
@@ -269,10 +269,8 @@ def run_stage(sessions, stage, test_block_config):
     )
 
     response = r.run()
-    logger.debug("Stage done")
 
     provider.end_tinctures(response)
-    logger.debug("tinctures done")
 
     for v in verifiers:
         saved = v.verify(response)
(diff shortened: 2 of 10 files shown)


michaelboulton:
https://github.com/taverntesting/tavern/pull/802

Issue #759, opened 2022-02-10
Provide a after_every_test_run hook


Art-S-D:
Currently my tests create a lot of dummy data
A hook that would run after every test would be usefull to cleanup everything my tests cretated

I tried to do it with a fixture but it is not convinient as it cannot access the variables dict

lvijnck:
+1 for this, would be highly relevant

michaelboulton:
Done in 1.20

PR #762, opened 2022-02-24
Add after test hook

lvijnck:
Solves #759 

I noticed the other PR or a more complicated mechanism to add before/after hooks. Though I really need this functionality as I'm running a large number of tests and I'm unable to perform proper cleanup now.

Happy to make changes as required.

### DIFF
--- a/tavern/testutils/pytesthook/item.py
+++ b/tavern/testutils/pytesthook/item.py
@@ -208,6 +208,12 @@ class YamlItem(pytest.Item):
         else:
             if xfail:
                 raise Exception("internal: xfail test did not fail '{}'".format(xfail))
+        # else:
+        #     if xfail:
+        #         logger.error("Expected test to fail")
+        #         raise exceptions.TestFailError(
+        #             "Expected test to fail at {} stage".format(xfail)
+        #         )
         finally:
             call_hook(
                 self.global_cfg,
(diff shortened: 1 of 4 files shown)


tavern/testutils/pytesthook/item.py
```diff
@@ -194,6 +194,16 @@ def runtest(self):
                     stage["name"] = stage["id"]
 
             run_test(self.path, self.spec, self.global_cfg)
+
+            # TODO: Will this also run after test failure?
```
michaelboulton, line 198: This won't be run if the test fails, you would need to put it in a 'finally' block or that to work.

michaelboulton (changes requested):
This is fine, but could you also just add a bit a documentation at the bottom of `docs/source/basics.md` near the other documentation for the hooks?

lvijnck:
> This is fine, but could you also just add a bit a documentation at the bottom of `docs/source/basics.md` near the other documentation for the hooks?

Done! LMK if any other changes are required.

P.S. I was checking whether there are any unit tests added for these hooks, but I could not directly find them.

michaelboulton:
> > This is fine, but could you also just add a bit a documentation at the bottom of `docs/source/basics.md` near the other documentation for the hooks?
> 
> Done! LMK if any other changes are required.
> 
> P.S. I was checking whether there are any unit tests added for these hooks, but I could not directly find them.

There aren't any tests for it as they only get run when you run a 'real test due to the way pytest works, if the integration tests pass then you can assume that the hook is being called correctly

michaelboulton (approved):

PR #783, opened 2022-06-03
Change Docker image tag

onbcst:
The Python-Alpine image does not work anymore without installing additional dependencies (python-dev). It is possible to use python-slim instead. Also updated to the latest version.


michaelboulton:
Thanks, I'm considering changing how all this works to avoid having to periodically update things like this

PR #802, opened 2022-07-30
tincture 2.0

michaelboulton:
#114

Add utility for running things before/after stages, able to be specified at test or stage level. Example functions

```python
def do_something_first(_):
    logger.info("starting stage!")


def time_request(_):
    t0 = time.time()
    yield
    t1 = time.time()
    logger.info("Request took %s", t1 - t0)


def print_response(_, extra_print="affa"):
    logger.info("STARTING:")
    r = yield
    logger.info("Response is %s (%s)", r, extra_print)
```
Used like
```yaml
---
test_name: Test tincture

tinctures:
- function: tavern.helpers:time_request

stages:
- name: do something
  tinctures:
  - function: tavern.helpers:print_response
    extra_kwargs:
      extra_print: "blooble"
  - function: tavern.helpers:do_something_first
  request:
    url: "{host}/echo"
    method: POST
  response:
    status_code: 200
```

Each stage takes the stage dictionary as its first argument. If the tincture is a generator (if it defines a 'yield' like the second example), then the response will be returned to it (if you want it). That's all they can be used for - there is no context saved between stages. The syntax for defining the functions/extra args is the same as existing external functions.

Tinctures are called when the stage is run, so _after_ any fixtures are evaluated.

As for the name - a [tincture](https://en.wikipedia.org/wiki/Tincture) is

> A tincture is typically an extract of plant or animal material dissolved in ethanol

I wanted a name that sounded sort of like 'fixture' and 'tavern'.


michaelboulton:
supersed by https://github.com/taverntesting/tavern/pull/844

PR #805, opened 2022-08-29
Use pyproject toml

michaelboulton:


michaelboulton:
Still just getting differing behaviour in CI and locally

Issue #816, opened 2022-10-25
Pytest no longer imports py lib


andrest50:
I use tavern for API interface testing, and I am getting this error `INTERNALERROR> AttributeError: module 'py' has no attribute 'io'`. Tavern gets the latest version of pytest (<8), and pytest just released 7.2.0, in which it removes dependency on `py`. As per the release (https://github.com/pytest-dev/pytest/releases/tag/7.2.0), you need to now directly install `py` library to make use of the `io` module - as it is still needed in this plugin.

IAmSilK:
Temporary fix for the time being that I have been using - install `py` alongside `tavern` so the package is available.

PR #818, opened 2022-11-02
Add py as dependency to remeditate AttributeError: module 'py' has no attribute 'io'

ahebrank:
See #816 .

pytest has removed py as a dependency as of 7.2.0 (https://github.com/pytest-dev/pytest/pull/10396), but tavern still requires it. Test failures produce an internal error `AttributeError: module 'py' has no attribute 'io'` without this dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,6 +28,7 @@ pyjwt>=2.4.0,<3
 paho-mqtt>=1.3.1,<=1.6.1
 jmespath<1
 pytest>=6.2,<8
+py
 python-box>4,<6
 stevedore
 
--- a/tox-integration.ini
+++ b/tox-integration.ini
@@ -17,6 +17,7 @@ changedir =
     generic: tests/integration
     noextra: tests/integration
 deps =
+    py
     docker-compose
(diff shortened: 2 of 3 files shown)


michaelboulton:
I think there is some issue with the way tox caching is working in the github action. I don't know why it keeps failing (due to it not _actually_ installing `py`) when it's explicitly listed as a dependency

michaelboulton (approved):

i540608:
@michaelboulton Hi, 
is it possible to remove the `Py` from the release dependencies and use different file for test dependencies, because the `Py` has [quite high-level security vulnerability](https://github.com/advisories/GHSA-w596-4wvx-j9j6) and it will trigger security scanners for my projects;

As the Py is itself deprecated, then it's highly likely that maintainers of the `Py` will not push the fix

michaelboulton:
It may be deprecated but pytest have also vendored _part_ of the library https://github.com/pytest-dev/pytest/pull/10396 , not the 'io' part which is also used by tavern. It means reaching into more internal pytest code.

PR #823, opened 2022-11-18
mqtt: fix unexpected behaviour

stintel:
Currently mqtt_response with the unexpected key set to true fails if no message is received. This is strange, was we don't expect a message, and should only fail if a message is received.

Formatted stage:
  mqtt_publish:
    payload: 10.10.10.10
    topic: inet6/add
  mqtt_response:
    payload: !anything ''
    timeout: 5
    topic: vallumd/will
    unexpected: true

Errors:
E   tavern.util.exceptions.TestFailError: Test 'add IPv4 IP to IPv6 ipset' failed:
    - Expected '<Tavern YAML sentinel for anything>' on topic 'vallumd/will' but no such message received

Fix this by suppressing the error MQTTResponse._await_response() and returning an empty dict if the unexpected key is set.

Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>

### DIFF
--- a/tavern/_plugins/mqtt/response.py
+++ b/tavern/_plugins/mqtt/response.py
@@ -181,11 +181,14 @@ class MQTTResponse(BaseResponse):
 
             self._maybe_run_validate_functions(msg)
         else:
-            self._adderr(
-                "Expected '%s' on topic '%s' but no such message received",
-                expected_payload,
-                topic,
-            )
+            if self.expected.get("unexpected"):
+                return {}
+            else:
+                self._adderr(
+                    "Expected '%s' on topic '%s' but no such message received",
+                    expected_payload,
+                    topic,
+                )
 
         if self.errors:
             if warnings:


tavern/_plugins/mqtt/response.py
```diff
@@ -181,11 +181,14 @@ def addwarning(w, *args, **kwargs):
...
-                topic,
-            )
+            if self.expected.get("unexpected"):
+                return {}
+            else:
+                self._adderr(
+                    "Expected '%s' on topic '%s' but no such message received",
+                    expected_payload,
+                    topic,
+                )
```
michaelboulton, line 191:
```suggestion
            if not self.expected.get("unexpected"):
                self._adderr(
                    "Expected '%s' on topic '%s' but no such message received",
                    expected_payload,
                    topic,
                )
```

Returning early will break the saving of variables from responses, this should do what is required

stintel: Indeed: [tavern.txt](https://github.com/taverntesting/tavern/files/10071620/tavern.txt)

stintel: My initial submission works, your suggestion including my extra attempts to fix it, doesn't. Any other ideas/suggestions?

michaelboulton: Sorry, I've been looking at the 'multiple mqtt responses' feature in Tavern 2.0 and wasn't thinking about there only being one mqtt response.

I think that what needs to be done is that my suggestion does need to be applied, but then also just after it checks if there were any errors (https://github.com/taverntesting/tavern/blob/540588f1a50adaccc1e16bd4032ba55d12233da5/tavern/_plugins/mqtt/response.py#L190-L197) it should do 
```python
if not msg:
    return {}
```
because there won't be any variables to save.

stintel: Made a slight alteration to your suggestion, seems to work like that. Let me know if that's OK for you?

michaelboulton:
I think the CI has randomly broken again, will try to fix it

stintel:
> I think the CI has randomly broken again, will try to fix it

Thanks, no rush, I can pip install tavern from my branch which allows me to continue working on the CI I'm using it in!

felixbohnus:
I also stumbled over this bug and tested the proposed solution with the local installation. I'm not sure about the processes: should I  create a new pull request or should this pull request be updated?

PR #839, opened 2023-01-27
Case insensitive matching

cazier:
I threw together this PR to support an issue I ran into trying to match specific words, case-insensitively. Tests pass on my machine! (Holy cow is the testing comprehensive and rigorous!)

Happy to make any changes, or address any comments.

Thanks!

### DIFF
--- a/docs/source/basics.md
+++ b/docs/source/basics.md
@@ -1557,6 +1557,31 @@ as in the example. If a named matching group is used as shown above, the saved v
       made_on: "2020-02-21"
 ```
 
+### Matching a string case-insensitively
+
+You might also have a specific string that you know the content will always be the
+same, but that can be in upper/lowercase. This can also be done using an external
+function (like regular expression matching), but as a shorthand there is also the
+`!anycase` YAML tag.
+
+```yaml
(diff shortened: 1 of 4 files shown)


michaelboulton:
This looks good but I'm wondering if this can't just be achieved using the existing [regular expression matching](https://tavern.readthedocs.io/en/latest/basics.html#matching-via-a-regular-expression) and [regex flags](https://docs.python.org/3/library/re.html#regular-expression-syntax), possible with some use of the ['A' flag](https://docs.python.org/3/library/re.html#re.A) eg

```yaml
  to_match: !re_match "(?i)ValUe"
```

From the docs of 'casefold' I'm not sure if this is the case for anything other than standard ascii characters though...

cazier:
I wasn't aware you could include the `(?i)` in-line with the regex pattern, so that definitely works for what I needed!

The casefold example *will* still fail, but I really only included it in the PR to be thorough, it's not something I need at the moment, and if it comes to it, could just use a union for it. (i.e., `!re_match "(?i)stra(?:ss|ß)e"`)

Thanks for taking a look, and ultimately fixing my issue!

Issue #863, opened 2023-05-09
verify_response_with not working on mqtt_response


octathorp:
Hello.

When calling an external function using "verify_response_with" simply nothing happens (with Tavern 2.0.7), but this was working perfectly with Tavern 1.25.2:

```
stages:
  - name: Send config
    mqtt_publish:
      topic: "{topic_config:s}"
      json: !include json/Config.json
    mqtt_response:
      topic: "{topic_config_ack:s}"
      json:
        debug: !anything
        status: 1
      timeout: 5
      verify_response_with:
        function: testing_utils:verifySetup
```
        
```
def verifySetup(response):
    with open("log.txt", "w") as log:
        log.write("test")
```
        
File "log.txt" is created when using Tavern 1.25.2 but nothing happens when using 2.0.7 nor 2.0.6.

At the same time, even if I try to refer to a non-existing function, the test simply passes, but with version 1.25.2 this was being checked in runtime:

`AttributeError: module 'testing_utils' has no attribute 'verifyNOTSetup'. Did you mean: 'verifySetup'?`

Is that a bug or something changed in most recent versions?

Thank you.

michaelboulton:
Release in 2.6.0

PR #864, opened 2023-05-12
#863 verify_response_with fix

octathorp:
This fixes issue #863 as stated in the comments.


tavern/response.py
```diff
@@ -131,7 +131,9 @@ def check_ext_functions(verify_block):
                     "Badly formatted 'verify_response_with' block"
                 )
 
-        check_ext_functions(response_block.get("verify_response_with", None))
+        mqtt_responses = response_block.get("mqtt_responses")
+        for mqtt_response in mqtt_responses:
+            check_ext_functions(mqtt_response.get("verify_response_with", None))
```
michaelboulton, line 136: This won't work because the response block can still be just a single mqtt response, or a rest response, or (soon) a gRPC response.

octathorp: As far as I tried with mqtt is working, with single or multiple responses, as even single responses reach this method in the json array form (with only one item, of course). You are totally right about rest response, simply I'm not using rest so I did not think about that.
No idea about future gRPC implementation, so no words about that.

I can try to update pull request with a proper rest response handling if that is enough, but I have no way to make a gRPC-future proof implementation. I was simply trying to fix the bug I reported some days ago.

michaelboulton: You don't need to worry about the grpc thing, I'll handle that in the appropriate PR, I mainly just meant it still needs to work for rest responses. Something like
```python
if mqtt_responses := response_block.get("mqtt_responses"):
  for mqtt_response in mqtt_responses:
    check_ext_functions(mqtt_response.get("verify_response_with", None))
else:
  check_ext_functions(response_block.get("verify_response_with", None))
```

octathorp: I have updated PR with your code suggestion, thank you!

michaelboulton (approved):

PR #870, opened 2023-05-27
Allow multi part file uploads

### LINKED ISSUE #833: YAML syntax prevents uploading multiple files with the same "name"
It looks like Tavern's current YAML syntax prevents using Tavern for certain standard kinds of multi-file uploads.

If you have an HTML form input like this:

```html
<input type="file" name="input_files" multiple="true" required="true"/>
```

Then the `POST` request should look like this:

```
POST /foo HTTP/1.1
User-Agent: PostmanRuntime/7.30.0
Accept: */*
Host: localhost:31950
Connection: keep-alive
Content-Type: multipart/form-data; boundary=--------------------------432665675269641474044671
Content-Length: 12345

----------------------------432665675269641474044671
Content-Disposition: form-data; name="input_files"; filename="file_1.txt"
Content-Type: application/octet-stream

first file conents
blah blah blah

----------------------------432665675269641474044671
Content-Disposition: form-data; name="input_files"; filename="file_2.txt"
Content-Type: application/octet-stream

second file contents
blah blah blah

----------------------------432665675269641474044671--
```

Notice how both files have `name="input_files"`, even though they have different `filename`s and different contents.

Quoting from [RFC 7578 section 4.3](https://datatracker.ietf.org/doc/html/rfc7578#section-4.3):

> The form data for a form field might include multiple files.
> ...
> To match widely deployed implementations, multiple files MUST be sent by supplying each file in a separate part but all with the same "name" parameter.

In `requests`, you can do this by [providing a list as the `files` argument](https://docs.python-requests.org/en/latest/user/advanced/#post-multiple-multipart-encoded-files):

```python
multiple_files = [
    ('input_files', ('file_1.txt', file_1, 'text/plain')),
    ('input_files', ('file_2.txt', file_1), 'text/plain'))]
r = requests.post(url, files=multiple_files)
```

But in Tavern's YAML syntax, this doesn't seem possible.

```yaml
request:
  url: '{url}'
  method: POST
  files:
    input_files: 'file_1.txt'
    input_files: 'file_2.txt' # Invalid YAML because this is a duplicate key. :(
```

michaelboulton:
Fixes #833

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -269,6 +269,12 @@ twine==4.0.2
     # via tavern (pyproject.toml)
 types-pyyaml==6.0.12.2
     # via tavern (pyproject.toml)
+types-requests==2.31.0.0
+    # via tavern (pyproject.toml)
+types-setuptools==67.8.0.0
+    # via tavern (pyproject.toml)
+types-urllib3==1.26.25.13
+    # via types-requests
 typing-extensions==4.4.0
     # via mypy
 urllib3==1.26.13
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,8 @@ dev = [
     "tox-travis",
     "twine",
     "wheel",
+    "types-setuptools",
+    "types-requests",
     # This has to be installed separately, otherwise you can't upload to pypi
     # "tbump@https://github.com/michaelboulton/tbump/archive/714ba8957a3c84b625608ceca39811ebe56229dc.zip",
 ]
@@ -132,6 +134,7 @@ addopts = [
     "--strict-markers",
     "-p", "no:logging",
     "--tb=short",
+    "--color=yes"
 ]
 norecursedirs = [
     ".git",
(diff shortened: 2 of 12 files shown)


SyntaxColoring:
Tested and confirmed that this works for my simple use case, which was like this:

```yaml
test_name: Upload multiple files

stages:
  - name: Upload multiple files
    request:
      url: 'http://localhost/foo'
      method: POST
      files:
        - form_field_name: files
          file_path: '/path/to/file/1'
        - form_field_name: files
          file_path: '/path/to/file/2'
    response:
      status_code: 201
```

Thank you!

broken-wheel:
Hi, just ran into an issue with this. I was trying to replicate
```
Content-Disposition: form-data; name="form1"; filename="file1"
```
which is equivalent to the example in `requests` (https://requests.readthedocs.io/en/latest/user/advanced/#post-multiple-multipart-encoded-files)
```
multiple_files = [
    ('images', ('foo.png', open('foo.png', 'rb'), 'image/png')),
    ('images', ('bar.png', open('bar.png', 'rb'), 'image/png'))]
```
but, there's no key to specify `filename` for each of the file and so the schema validation fails. How do I get around this?

PR #922, opened 2024-03-28
Replace deprecated `pytest.warns(None)`

mgorny:
Replace the deprecated `pytest.warns(None)` use with the `error` warning filter, as recommended
in https://github.com/pytest-dev/pytest/issues/9404#issue-1076710891. With this change, the tests pass with pytest-8.1.1.


mgorny:
Alternatively, I can change it to use `warnings.catch_warnings(record=True)` or the `recwarn` fixture, and continue to check for zero recorded warnings.

michaelboulton:
> Alternatively, I can change it to use `warnings.catch_warnings(record=True)` or the `recwarn` fixture, and continue to check for zero recorded warnings.

I think it's best to do this

mgorny:
I'm sorry, which one?

michaelboulton:
> I'm sorry, which one?

Probably the first one, seems like that makes the most sense.

mgorny:
I've updated the pull request now.

michaelboulton (approved):

mgorny:
Thanks!

PR #955, opened 2025-03-26
feat: add support for anynumber

e-kazakov:
## Problem
JSON does not differentiate natively between floats and ints. Depending on how the server formats the response for a number, we may receive `23` and `23.4` in the same field
Whole value
```json
{
    "a_number_field": 23
}
``` 

Fractional value
```json
{
    "a_number_field": 23.4
}
```

Existing type matchers `!anyint` and `!anyfloat` do not work for a test that has no control over the exact value in the response. Because `!anyfloat` does not accept a whole number in response, failing with error (_formatting is mine for better readability_):
```
tavern._core.exceptions.KeyMismatchError: Type of returned data was different than expected (
  expected["a_number_field"] = '<Tavern YAML sentinel for <class 'float'>>' (
    type = <class 'tavern._core.loader.FloatSentinel'>
  ),
  actual["a_number_field"] = '23' (type = <class 'int'>)
)
```
That forces us to use `!anything`, which is less restrictive.

## Solution
This PR introduces a new `!anynumber` type matcher that would match both ints and floats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# UNRELEASED
+
+- Add !anynumber type sentinel for matching both integers and floats in responses
+
 ##  0.1.2           Allow sending/validation of JSON lists (2017-11-21)
 
 ##  0.1.3           Fix global configuration loading via pytest command line (2017-12-05)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # Changelog
 
-# UNRELEASED
-
-- Add !anynumber type sentinel for matching both integers and floats in responses
-
 ##  0.1.2           Allow sending/validation of JSON lists (2017-11-21)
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+# UNRELEASED
+
+- Add !anynumber type sentinel for matching both integers and floats in responses
+
```
michaelboulton, line 6: This can be removed

michaelboulton (commented):
Good idea, look nice and clean, just one minor comment about the changelog

CHANGELOG.md
e-kazakov: Removed

e-kazakov:
Hey @michaelboulton, is there anything else you'd like me to address?

michaelboulton:
Well I don't know why the CI is failing but it looks like another github error :shrug: 

Thanks for the contribution, I'll release it shortly

PR #966, opened 2025-07-01
feature: add support for marks with args

ChristophSchwenk:


michaelboulton:
The mqtt error is unrelated due to using an old version of debian, I'll fix that separately.

Thanks for the contribution, I'll release this ASAP

PR #1024, opened 2026-04-18
Handle optional dependencies gracefully in tests and plugins

pyd-07:
## Summary

Fixes test collection failures caused by missing optional dependencies such as gql, grpc, paho, faker, hypothesis, and pydantic.

Previously, running `pytest` in a clean environment would fail during collection due to ImportError exceptions raised by plugin modules and test files importing optional dependencies at module level.

---

### 1. Test Suite Improvements
- Added `pytest.importorskip()` in test files that depend on optional libraries:
  - gql (GraphQL)
  - grpc (gRPC)
  - paho (MQTT)
  - faker, hypothesis, pydantic (test/dev dependencies)
- Ensures tests are skipped instead of failing when dependencies are not installed

### 2. Plugin Import Safety
- Wrapped optional dependency imports in `try/except ImportError` blocks across:
  - GraphQL plugin
  - gRPC plugin
  - MQTT plugin
- Prevents ImportError during pytest collection

### 3. Type Annotation Fixes
- Replaced direct type annotations with string annotations for optional dependency types
- Avoids runtime evaluation issues when dependencies are missing

### 4. gRPC Status Validation Fallback
- Updated `to_grpc_status()` to handle absence of grpc gracefully
- Added fallback validation for:
  - string-based status names (e.g. "UNAVAILABLE")
  - integer status codes (0–16)
- Ensures consistent validation behavior even when grpc is not installed

---

## Result

- pytest runs successfully without requiring all optional dependencies
- Relevant tests are skipped instead of failing
- No changes to core functionality or behavior

Example result in a clean environment:

## Release Notes

* **Bug Fixes**
  * Improved handling of optional dependencies to ensure the application loads and functions gracefully when extensions are not installed.

* **Tests**
  * Updated test suite to skip tests gracefully when optional dependencies are unavailable, improving test execution reliability.


michaelboulton:
To explain why this specific approach doesn't work, using pytest.importorskip inside production code means that user's tests might randomly be skipped with no obvious feedback why. Production code should never rely on test utilities to handle imports.

If a user doesn't have the dependencies installed, it failing is the correct behaviour. If a required dependency is missing, the code should fail explicitly so the user knows they need to install it. Because a user will install tavern from pypi, this will automatically install all the dependencies for them. If running tests locally, then the tests should be run with `tox` which also installs all the dependencies for you.

Thanks for making a contribution but I'm going to close this PR for the above reasons.

pyd-07:
> To explain why this specific approach doesn't work, using pytest.importorskip inside production code means that user's tests might randomly be skipped with no obvious feedback why. Production code should never rely on test utilities to handle imports.
> 
> If a user doesn't have the dependencies installed, it failing is the correct behaviour. If a required dependency is missing, the code should fail explicitly so the user knows they need to install it. Because a user will install tavern from pypi, this will automatically install all the dependencies for them. If running tests locally, then the tests should be run with `tox` which also installs all the dependencies for you.
> 
> Thanks for making a contribution but I'm going to close this PR for the above reasons.

Thanks for the feedback, that makes sense regarding separation between test utilities and production code.

I understand the concern about relying on pytest.importorskip and altering expected dependency behavior. I’ll keep this in mind for future contributions.

Appreciate you taking the time to review!

Issue #733, opened 2021-11-11
Parsing Error


KenStorey:
`platform linux -- Python 3.9.0, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /usr/local/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.9.0', 'Platform': 'Linux-4.19.128-microsoft-standard-x86_64-with', 'Packages': {'pytest': '6.2.5', 'py': '1.11.0', 
'pluggy': '1.0.0'}, 'Plugins': {'metadata': '1.11.0', 'tavern': '1.17.0', 'html': '3.1.1'}}
rootdir: /tavern_tests
plugins: metadata-1.11.0, tavern-1.17.0, html-3.1.1
`

I have the following stage
`

     mqtt_publish:
      topic: "{hmi_display_field_topic}"
      json:
        application: "{test_app}"
        application_ref: 5000
        field_list:
          - "registered-identity"
          - "func-alias"
          - "status-1"
          - "action"
          - "explanation-1"
          - "status-2"
          - "status-3"
          - "status-4"
        field_values:
          - "Registered Identity"
          - "Alias"
          - "Status 1"
          - "Action"
          - "Explanation 1"
          - "S#2"
          - "S#3"
          - "Status#4"
        field_justification_values: [0, 0, 0, 0, 0, 0, 0, 0]
        field_flashing_values: [false, false, false, false, false, false, false, false]
        field_highlighted_values: [false, false, false, false, false, false, false, false]
`
Which validates correctly however, on execution, Tavern reports

`FAILED tavern_tests/test_tdcp_display_field.tavern.yaml::Multiple fields - ValueError: expected '}' before end of string`

I have attached a full stack trace.

[tavern_stack_trace.txt](https://github.com/taverntesting/tavern/files/7522342/tavern_stack_trace.txt)

michaelboulton:
It looks like this is because the thing it's trying to format:
```
to_format = '{Explanation 1      '
```
Has an opening curly brace but no closing curly brace which is making the [Formatter.parse](https://docs.python.org/3/library/string.html#string.Formatter.parse) method throw an error.

As this isn't controlled by Tavern, a good solution might be to log a warning if this happens and just not format the given string

PR #1031, opened 2026-04-29
docs: add note about unmatched braces in format strings

devimano2011:
Related to #733

The existing docs explain how to escape literal `{` and `}` using `{{` and `}}`. However they don't mention what happens when a string contains an **unmatched** brace (e.g. `{Explanation 1` from an MQTT payload), which previously caused an unhandled `ValueError`.

Adds a short `**NOTE**` paragraph after the existing escaping example to explain that strings with unmatched braces will log a warning and be passed through unformatted rather than raising an error, and directs users to escape literal braces to avoid unexpected behaviour.

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation now clarifies string formatting behaviour when encountering unmatched curly braces. The guidance specifies that Tavern will emit a warning and pass strings through without formatting, rather than failing. Comprehensive guidance provided on escaping literal curly braces to prevent unexpected formatting results.

### DIFF
--- a/docs/source/basics.md
+++ b/docs/source/basics.md
@@ -204,14 +204,16 @@ request:
     graphql_query: "{%raw%}{{ user(id: 123) {{ first_name }} }}{%endraw%}"
 ```
 
-**NOTE**: This is a toy example, if you want to use GraphQL with tavern then 
+**NOTE**: This is a toy example, if you want to use GraphQL with tavern then
 head over to the [GraphQL docs](./graphql.md).
 
 **NOTE**: If a string value contains an unmatched `{` that is not a valid
-format placeholder (for example, an MQTT payload value like `{Explanation 1`),
-Tavern will log a warning and pass the string through unformatted rather than
-raising an error. To avoid unexpected behaviour, always escape literal curly
-braces as described above.
+format placeholder, Tavern will log a warning and pass the string through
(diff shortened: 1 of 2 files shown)


docs/source/basics.md
```diff
@@ -207,6 +207,12 @@ request:
 **NOTE**: This is a toy example, if you want to use GraphQL with tavern then 
 head over to the [GraphQL docs](./graphql.md).
 
+**NOTE**: If a string value contains an unmatched `{` that is not a valid
+format placeholder (for example, an MQTT payload value like `{Explanation 1`),
+Tavern will log a warning and pass the string through unformatted rather than
+raising an error. To avoid unexpected behaviour, always escape literal curly
+braces as described above.
```
michaelboulton: This is not true.

michaelboulton (changes requested):

devimano2011:
Hi @michaelboulton, thank you for the review!

You were correct. Fixed in the latest commit: `formatter.parse()` returns a lazy generator, so iterating it raised `ValueError` for unmatched braces before reaching the existing try/except. The fix wraps the parse call in `list()` inside its own try/except ValueError so unmatched braces are caught early and the string is passed through unformatted with a warning - matching what the docs describe.

No changes needed to `basics.md` as it already accurately describes the intended behaviour. Please let me know if anything else needs addressing!

PR #1033, opened 2026-05-01
feat: add global tinctures support via tavern-global-cfg (fixes #969)

### LINKED ISSUE #969: How to apply tinctures to all tests?
[This example](https://tavern.readthedocs.io/en/latest/basics.html#tinctures) in the documentation is awesome:

```py
def time_request(stage):
    t0 = time.time()
    yield
    t1 = time.time()
    logger.info("Request for stage %s took %s", stage, t1 - t0)
```

```yaml
# test_stuff.tavern.yaml
tinctures:
  - function: package.helpers:time_request
```

But I want to apply this to all tests and get all requests performance measured, so instead of adding these two lines of code to every single test, I'd like to use the [including external files](https://tavern.readthedocs.io/en/latest/basics.html#including-external-files) feature. I created an `yaml` file like this:

```yaml
# include_tinctures.yaml
tinctures:
  - function: package.helpers:time_request
```

Then added this to `pytest.ini`:

```cfg
[pytest]
tavern-global-cfg =
    include_tinctures.yaml
```

This should work like if we had this in every test, right?

```yaml
includes:
  - !include include_tinctures.yaml
```

But I'm unable to make it work. Am I missing anything? Being able to do so would be awesome, tavern would become a performance tester too.

Thank you for the awesome tool by the way!!

devimano2011:
Fixes #969

## Problem
Tinctures could only be applied per-test/stage. Users could not apply a tincture globally to all tests via `tavern-global-cfg`, despite variables and stages already supporting this.

## Solution
This PR adds global tinctures support across 3 files:

1. **`tavern/_core/pytest/config.py`** - Added `tinctures: list` field to `TestConfig` dataclass
2. 2. **`tavern/_core/pytest/util.py`** - Loads `tinctures` from global config files (same as `stages` and `variables`)
3. 3. **`tavern/_core/run.py`** - Merges global tinctures into stage tinctures in `run_stage()`
## Usage
Users can now add global tinctures in their `tavern-global-cfg` file:

```yaml
# global_config.yaml
tinctures:
  - function: mypackage.helpers:time_request
  - ```
```ini
# pytest.ini
[pytest]
tavern-global-cfg = global_config.yaml
```

## Summary by CodeRabbit

* **New Features**
  * Test configuration now supports global tinctures settings that automatically merge with stage-specific tinctures, enabling centralised configuration management.


tavern/_core/run.py
```diff
@@ -340,6 +340,10 @@ class _TestRunner:
 
     def run_stage(self, idx: int, stage, *, is_final: bool = False) -> None:
         tinctures = get_stage_tinctures(stage, self.test_spec)
+        # Merge global tinctures from test config (fixes #969)
+        from tavern._core.extfunctions import get_wrapped_response_function
```
michaelboulton: Make this import global and remove the comment

michaelboulton (changes requested):
Please also:
- Add some unit tests (eg, that call run_test and ensure that the 'global' tinctures are called)
- Add integration tests if possible (though this might be a bit difficult to do?)
- Update the docs in external_code.md

devimano2011:
Fixed the CI failure. The root cause was that `get_wrapped_response_function` was imported inside the `run_stage` method as a local import, so `patch("tavern._core.run.get_wrapped_response_function", ...)` in the unit tests had no module-level name to patch against and the mock was never applied.

Fix: moved `from .extfunctions import get_wrapped_response_function` to the top-level imports in `run.py` and removed the local import from inside the method. The unit tests should now pass.

PR #1041, opened 2026-05-16
fix(rest): restore session cookies after per-request cookie override even on exception

reachsridhard:
### Summary

 This PR ensures that session cookies are correctly restored after a per-request cookie override, even if an exception occurs during the request execution.

### Issue

 In the REST plugin, when a specific request defines its own cookies , Tavern temporarily clears the session's cookie jar to ensure only the specified cookies are sent. Previously, this restoration logic in **_set_cookies_for_request** did not use a try...finally block. If an exception (such as a **ConnectionError , Timeout , or any other runtime error)** was raised during the request, the code execution would bypass the restoration step, leaving the session with an empty cookie jar for subsequent stages.

Solution Wrapped the **yield statement in the _set_cookies_for_request** context manager within a **try...finally block.** This guarantees that the original session cookies are restored regardless of whether the request succeeds or fails with an error.

### Changes

- request.py
  - Updated _set_cookies_for_request to use try...finally for cookie restoration.
  - Removed the outdated docstring comment that claimed cookie restoration was unnecessary on failure.
- test_request.py
  - Added test_cookies_restored_after_request_exception to verify that cookies are correctly restored when an exception is raised inside the context manager.
### Verification Results

- New Test Case : test_cookies_restored_after_request_exception specifically simulates an exception during a request and asserts that the session cookies are correctly restored to their pre-request state.
- Verified that normal execution (no exception) still restores cookies as expected.


michaelboulton (commented):
I think that because this was left unspecified before - can you add a note into the documentation to note that (After this change) any 'finally' blocks will use any cookies that have been accumulated up to the stage which failed?
ie if there is
- stage 1 (gets cookies)
- stage 2 (fails)
- stage 3
- finally block
Then the 'finally' block will use the cookies from stage 1. The current behaviour on main means that the cookies will not be present (I think?)

reachsridhard:
> I think that because this was left unspecified before - can you add a note into the documentation to note that (After this change) any 'finally' blocks will use any cookies that have been accumulated up to the stage which failed? ie if there is
> 
> * stage 1 (gets cookies)
> * stage 2 (fails)
> * stage 3
> * finally block
>   Then the 'finally' block will use the cookies from stage 1. The current behaviour on main means that the cookies will not be present (I think?)

@michaelboulton  I’ve added a note to the docs to make this explicit.

Location: docs/source/core_concepts/flow.md → “Finalising stages”
Summary of note:
Finally stages run using the same persistent HTTP session as the rest of the test.
If a test fails mid-way, any cookies accumulated up to the failing stage are preserved and will be sent with requests made in the finally block.
Example called out: stage 1 (sets cookies) → stage 2 (fails) → finally (cleanup) runs with cookies from stage 1.

reachsridhard:
@michaelboulton If you’d like, I can mirror a shorter reminder in docs/source/http.md under “Persistent cookies” as well.

michaelboulton (approved):

PR #1065, opened 2026-06-28
feat: replace check_expected_keys with pydantic models (#1034)

### LINKED ISSUE #1034: Replace 'check_expected_keys' with pydantic or something
This was added about 9 years ago and there are more modern ways to do it

reachsridhard:
Replace the 9-year-old check_expected_keys pattern with pydantic models that use extra='forbid' for key validation. This provides the same behavior (rejecting unexpected keys) with better error messages and type safety.

Changes:
- Add tavern/_core/pydantic_models.py with pydantic models for each request/client spec (REST, MQTT, gRPC)
- Replace all check_expected_keys calls with .validate_keys() calls in rest/request.py, mqtt/request.py, mqtt/client.py, grpc/request.py, grpc/response.py, grpc/client.py
- Remove now-unused expected_blocks dicts from MQTT and gRPC clients
- Add tests/unit/test_pydantic_models.py with 27 tests covering all new models
- Keep check_expected_keys in dict_util.py for backward compatibility

Closes #1034

## Summary by CodeRabbit

* **New Features**
  * Added stricter validation for REST, MQTT, and gRPC request/client settings to reject unexpected fields.
  * Improved handling of nested configuration blocks, including TLS and connection settings.

* **Bug Fixes**
  * Validation errors now return clearer messages when invalid keys are supplied.

* **Tests**
  * Added coverage for valid and invalid configuration inputs across REST, MQTT, and gRPC.


tavern/_core/pydantic_models.py
```diff
@@ -0,0 +1,159 @@
...
+    auth: Optional[Any] = None
+    json_body: Optional[Any] = Field(default=None, alias="json")
+    verify: Optional[Any] = None
+    files: Optional[Any] = None
+    file_body: Optional[Any] = None
+    stream: Optional[Any] = None
+    timeout: Optional[Any] = None
+    cookies: Optional[Any] = None
+    cert: Optional[Any] = None
+    follow_redirects: Optional[Any] = None
```
michaelboulton, line 67: One (fairly big) note is that having these as  'Any' defeats the point of having pydantic which does have a lot of type checking built into it, otherwise just using dataclasses/dacite would be just as effective.

michaelboulton (changes requested):
The implementation in general is sold but adding pydantic should also have type checking with it.

PR #1072, opened 2026-07-27
docker: run example and test containers as non-root user

chernojagne:
Add a non-root `USER` directive to the four Dockerfiles used by Tavern's examples and integration tests.

**What changes**

Each of `example/graphql/Dockerfile`, `example/grpc/Dockerfile`, `example/http/Dockerfile`, and `tests/integration/Dockerfile` now creates a system user (`app`, UID 1001) and drops privileges before the `CMD` instruction.

**Why**

Running containers as root is a well-known security risk. The numeric UID (`USER 1001`) also satisfies Kubernetes security policies that enforce `runAsNonRoot` and a specific `runAsUser` value, so these example containers can be used as-is inside restricted clusters.

No application behaviour is changed — only the effective UID inside the container.

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.

## Summary by CodeRabbit

- **Security**
  - Example and integration containers now run as a dedicated non-root user by default, improving runtime isolation and reducing security risk.
- **Compatibility**
  - Existing container build processes and startup commands remain unchanged.

### DIFF
--- a/example/graphql/Dockerfile
+++ b/example/graphql/Dockerfile
@@ -9,4 +9,8 @@ COPY . /app
 RUN uv sync
 
 ENV PYTHONPATH=/app/
+
+# Run as a non-root user for container security
+RUN groupadd --system app && useradd --system --uid 1001 --gid app --create-home app
+USER 1001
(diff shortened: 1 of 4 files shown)


michaelboulton (approved):

michaelboulton:
:+1:

PR #1073, opened 2026-07-31
feat(starlark): Add per-stage 'if' and 'retry_until' expressions

michaelboulton:
add gha style 'if' and 'retry until' actions. reuses the same starlark code.


docs/source/scripting.md
```diff
@@ -66,6 +74,183 @@ Starlark control flow is an experimental feature. Enable it with the pytest flag
...
+the same embedded interpreter. There is no script, no `load()`, and stages do not need an `id`. These need the same
+`--tavern-experimental-starlark-pipeline` flag as `control_flow`.
+
+All test variables - anything from `save`, `includes`, global config, fixtures, parametrisation, and the `tavern` box -
+are bound directly as Starlark globals.
+
+> **Important:** unlike almost everywhere else in Tavern, these expressions are **not** format-string interpolated.
+> Write `if: var_x > 2`, not `if: "{var_x} > 2"` - the latter compares the literal string `"{var_x}"` and will silently
+> do the wrong thing. Any variable whose name is not a valid Starlark identifier (for example one containing a dash) is
+> not bound.
```
michaelboulton: maybe change this to use formatting so its consistent with existing formatting checks, and to allow vars with dashes

michaelboulton: eg if
```python
vars = {"my-thing": 2}
```
then
```yaml
if: "{my-thing} > 4"
```
converted internally to
```yaml
if: "2 > 4"
# Error message include the original format string too
```
Or maybe something hardcoded like
```yaml
# 'vars' injected as a global variable
if: vars.get("my-thing") > 4
```
(downside with the second one is that it prevents doing something like "{my_int!.2f}"

michaelboulton: Or, even force users to do 
```yaml
if: vars["my-thing"] > 4
```
explicit 'vars' global variable, like response

tavern/_core/run.py
```diff
@@ -303,6 +304,15 @@ def getonly(stage):
         try:
             # Run tests in a path in order
             for idx, stage in enumerate(test_spec["stages"]):
+                if "skip" in stage:
+                    warnings.warn(  # noqa
+                        f"Stage '{stage['name']}' uses the 'skip' key, which is deprecated - "
+                        "use the 'if' key instead (note that the logic is inverted, and it uses "
+                        "Starlark rather than simpleeval). See 'Running a stage conditionally "
+                        "with if' in the scripting documentation.",
+                        DeprecationWarning,
```
michaelboulton: Remove warning until the 'if' statement works in mqtt/grpc/etc as well

tavern/_core/schema/tests.jsonschema.yaml
```diff
@@ -111,13 +122,33 @@ definitions:
         default: 0
 
       skip:
+        deprecated: true
         oneOf:
           - type: boolean
-            description: Whether to skip this stage
+            description: Deprecated, use 'if' - whether to skip this stage
```
michaelboulton: remove deprecation

